PR #2, opened 2018-08-13
Coding test submission

focusaurus:
Here is an implementation of the REST API to complete the coding test.

- Developer docs are in `developer-guide.md`
- Implemented on hapi and mongodb
- Includes automated tests
  - Tests run as local tests by default but can also target a remote server
  - The tests are mostly not strict/pure unit tests. Since we're dockerized they do really interact with mongodb and are triggered mostly via the actual HTTP endpoints. However, the same tooling supports strict unit tests and testing individual functions directly as well.
- Includes code coverage tooling
- Includes eslint static analysis tooling and autoformatting tooling
- Includes docker-compose setup
- Includes structured logging
- API documentation via openapi/swagger served at `http://localhost:3000/documentation` when the server is running (hapi-swagger npm package)
- Input validation and API shapes modeled as joi schemas
- Server configured via environment variables and includes validation of config
- Commit messages should be satisfactory


developer-guide.md
```diff
@@ -0,0 +1,38 @@
...
+  - `node --inspect=0.0.0.0:9229 .`
+  - or if you need an immediate breakpoint: `node --inspect-brk=0.0.0.0:9229 .`
+  - Then in chrome browser, browse to `chrome://inspect`
+- Run all the unit tests: `npm test`
+- Run a single unit test file: `tap code/some-test-tap.js`
+- Run a group of unit test files: `tap 'code/some-glob-*-tap.js'`
+- Copy some data out of the running mongodb container to your dev machine
+  - `docker-compose exec mongo bash`
+  - `cp /path/to/some/file.txt /host`
+  - The file will be at `datat/host/file.txt` under this repo's working directory
```
wprater: My turn to find a typo. `datat/host/file.txt` should be `data/host/file.txt`

docker-compose.yml
```diff
@@ -0,0 +1,24 @@
...
+      - '27017'
+    # /host for ad-hoc transfers in/out of container
+    # /data for DB data itself
+    volumes:
+      - './data/host:/host'
+      - './data/mongo:/data'
+
+  hapi:
+    build: .
+    command: node --inspect=[::]:9229 code/main.js
```
wprater, line 16: Glad you're listening to the loopback port here, otherwise you can get debugging issues.  I was able to debug this app with VS Code Node debugger.

code/users/create-tap.js
```diff
@@ -9,8 +10,8 @@ tap.beforeEach(async () => {
 
 tap.test("create user", test => {
   const user = {
-    id: "f47ac10b-58cc-4372-a567-0e02b2c3d479",
-    name: "Pat Patterson"
+    id: uuid(),
```
wprater: curious why you chose to require a client to supple the UUID vs the system itself?

package.json
```diff
@@ -9,9 +9,11 @@
     "hapi": "17.5.3",
     "hapi-mongodb": "7.1.0",
     "hapi-pino": "5.0.1",
```
wprater: Id never used this logger before.  Thanks for pointing it out.

code/users/plugin.js
```diff
@@ -50,9 +51,43 @@ module.exports = {
         if (!user) {
           throw boom.notFound();
         }
-        return h.response(fromMongo(user));
+        return h.response(modelUtils.fromMongo(user));
       }
     });
+
+    server.route({
+      method: "PATCH",
```
wprater: Nice to see patch vs put for partial updates.

code/users/plugin.js
```diff
@@ -0,0 +1,113 @@
...
+      method: "POST",
+      path: "/",
+      options: {
+        tags: ["api"],
+        response: { schema: schema.get },
+        validate: {
+          payload: schema.create
+        }
+      },
+      handler: async (request, h) => {
```
wprater, line 19: What do you think about the concept of separating these handlers into another export, so they're not all anonymous functions?

wprater (commented):
Great work.  Nice progression through your commit messages :)

I was not able to get this running per your instructions.  The docker container is running as a user `501` which does not exist, so NPM needs to have its cache location reconfigured.  `$HOME` may also need to be changed to some world writable location. 

Added some notes.  No more work is required, but please comment and feel free to address the changes if you're up for it. 

cc @codin-sf @aaron-sf

package.json
```diff
@@ -0,0 +1,58 @@
...
+    "node-dev": "3.1.3",
+    "prettier": "1.14.2",
+    "prettier-eslint-cli": "4.7.1",
+    "supertest": "3.1.0",
+    "tap": "12.0.1"
+  },
+  "scripts": {
+    "test": "tap 'code/**/*-tap.js'",
+    "coverage": "tap --jobs-auto --coverage-report=html 'code/**/*-tap.js'",
+    "format": "prettier-eslint --write 'code/**/*.js'",
```
wprater, line 37: Would be nice to run prettier on pre-commit and staged files using husky or a technique like this https://gist.github.com/primaryobjects/9ab8d2346aa58ef6959ad357bb835963

wprater: This way there will be little to no chance of poorly formatting code making it's way into the repo.

focusaurus:
So with the non-root docker setup I got lucky in that both my host OS and the container OS were ubuntu and my uids were both 1000 by coincidence. I just pushed a build script that will ensure the container has a matching user account such that editing files from the host and the container should play nicely together. Tested on macOS and I was able to get the tests passing and the app running in the container.

code/users/plugin.js
focusaurus: It's a fine concept, but for tiny CRUD handlers I actually find the maintenance much easier when I can grep the code for the API path and edit that file and be done. If the handlers get larger, sure split them out to separate named functions or even separate modules, but I start with defining them inline. My experience with projects where everything is split out too much is navigating around to the right file gets tedious.

package.json
focusaurus: I'd be OK with that but I'd rather have it fail and abort the commit rather than running code formatting during a git hook and proceeding with a commit. I have found you can go too far with git hooks which tend to be project-specific custom behavior and git is a common tool that should behave mostly consistent across all projects in my opinion. I find it very jarring if I type a git command that does X in 99 projects and in this project it does X plus Y where Y is a very different thing in my mind (like pushing a branch or changing the source code). Thus I tend to not have nor want git hooks that alter source code. I leave the source code changing to the formatting script and git for committing the code as is.

code/users/plugin.js
wprater: Makes perfect sense.

package.json
wprater: We can write contribution guidelines that can address this need to.  Many editors can run prettier and eslint on demand.

Thanks for your insights.

wprater:
@focusaurus well done.  Thank you for your contribution, we can close this now and I will reach out to you.  Thanks again for your time here, we very thorough and executed with care.

PR #3, opened 2018-08-28
API complete w/ extra features

pkwu:
- create/read/update/delete
- MongoDB + mongoose
- API documentation available in /docs/rest-server.md
- Documentation on how to run code available in /docs/README.md
- Testing available via mocha/chai, documentation available in /docs/test.md
- Read only endpoint to fetch location information based off the user's address using Mapbox API, documentation available in /docs/rest-server.md (API KEY required)
- Docker Compose configured, documentation on how to run docker available in /docs/docker.md


bin/buildEnv.js
```diff
@@ -0,0 +1,19 @@
...
+}
+
+const buildEnv = () => {
+  _.each(envVariables, (value, key) => {
+    fs.writeFileSync(`./${key}/.env`, '')
+    createENVFile(key, value);
+  });
+}
+
+buildEnv();
```
wprater, line 19: You should consider allowing this to be executed as a script by setting the executable flag and adding a exec reference to the top of the file like such `#!/usr/bin/env node`

However, better practice is to assume the ENV will be setup in the environment running the app and only using `dotenv` in development only. 

You should instead ensure that all requires ENV properties are set vs. looking for a `.env` file to be present.

rest-server/test/user.js
```diff
@@ -0,0 +1,257 @@
...
+      const newUser = new User({
+        name: 'Test User',
+        dob: '08/27/2018',
+        address: 'Los Angeles',
+        description: 'This is a test'
+      });
+      newUser.save( (err, data) => {
+        chai.request(server)
+          .get(`/api/users/${data.id}`)
+          .end( (err, res) => {
```
codin-sf, line 47: Chai is pretty cool because it has some plugins to extend base functionality. https://github.com/chaijs/deep-eql allows you to compare a complete structure without breaking each node out for an assertion.

bin/buildEnv.js
```diff
@@ -0,0 +1,19 @@
+const fs = require('fs');
+const _ = require('lodash');
```
wprater: Curious as to why you've chosen to import `lodash` in it's entirety vs. the each module being used instead?  Is there a reason for using `_.each` vs. the native `Array#forEach`?

docs/rest-server.md
```diff
@@ -0,0 +1,44 @@
+Database Model + Attributes
+Example:
+### MODEL NAME
+* Attribute Name: data type & (additional information)
+
+### USER
```
wprater, line 6: Is there an express plugin to generate and OpenApi swagger file automatically?  Might be worth looking into.

package.json
```diff
@@ -0,0 +1,41 @@
...
+    "url": "https://github.com/pkwu/node-backend-test/issues"
+  },
+  "homepage": "https://github.com/pkwu/node-backend-test#readme",
+  "devDependencies": {
+    "babel-polyfill": "^6.26.0",
+    "babel-preset-env": "^1.7.0",
+    "babel-register": "^6.26.0",
+    "chai": "^4.1.2",
+    "chai-http": "^4.0.0",
+    "cors": "^2.8.4",
```
wprater: I don't think you want to have the following as dev dependencies:

- cors
- express
- helmet

rest-server/src/components/mapbox/mapboxControllers.js
```diff
@@ -0,0 +1,19 @@
+import axios from 'axios';
+import { fetchSingleUserHelper } from '../users/userHelpers';
+
+const mapboxAPIKey = process.env.mapboxAPIKey || 'MUST-REPLACE-WITH-APIKEY-TO-PASS-TEST';
```
codin-sf: You can even add the mapboxAPIKey value into your Dockerfile.

rest-server/src/components/mapbox/mapboxControllers.js
```diff
@@ -0,0 +1,19 @@
...
+
+const fetchUserLocationCoordinates = async (req, res) => {
+  try {
+    const singleUser = await fetchSingleUserHelper(req.params);
+    const loc = singleUser.address.split(' ').join('%20');
```
codin-sf: When constructing URLs you can encode the string entirely if you need to (https://www.w3schools.com/jsref/jsref_encodeURI.asp).

rest-server/bin/entry.js
```diff
@@ -0,0 +1,17 @@
+require('babel-register');
```
wprater, line 1: You really should not be using `babel-register` in a production app.  If you need ECMA features not available it `LTS`, a transpilation build step should be used.  However, what babel features do you need?

rest-server/bin/entry.js
```diff
@@ -0,0 +1,17 @@
...
+require('babel-polyfill');
+
+const fs = require('fs');
+const path = require('path');
+const env = require('dotenv');
+
+if (!fs.readFileSync(path.resolve(__dirname, '../.env'))) {
```
wprater: In `production` it should not be assume the `.env` file will be present.  Instead, you should require the needed ENV variables are present, and if not, then fail.  Don't fail on the lack of a present `.env` file.

rest-server/src/components/users/userControllers.js
```diff
@@ -0,0 +1,65 @@
...
+  try{
+    const singleUser = await fetchSingleUserHelper(req.params);
+    res.status(200).send(singleUser);
+  }
+  catch (err) {
+    res.status(400).send({error: 'Could not find User'});
+  }
+};
+
+const updateUser = async (req, res) => {
```
codin-sf, line 33: This method is a bit confusing. It looks like you are trying to combine PUT (or POST) and a PATCH. Meaning, you are trying to add a user as well as update a user based on fragmented data and not a complete object.

This could probably use some additional thought and abstraction into separate handlers.

rest-server/src/config/express/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+import router from '../../routes/index';
+
+const middleWare = [
+  helmet(),
+  parser.json(),
+  parser.urlencoded({ extended: true }),
+  cors({
+    allowedHeaders: 'Content-Type, authorization',
+    methods: ['GET, POST, PUT, DELETE', 'OPTIONS'],
```
codin-sf: Be careful here. You do not support OPTIONS.

codin-sf (commented):
Overall you have some great work here. I can tell by the commit history and your implementation that you have put a lot of thought and effort in. There is some room for improvement but you have done an excellent job.

rest-server/src/components/mapbox/mapboxControllers.js
```diff
@@ -0,0 +1,19 @@
...
+    const url = `https://api.mapbox.com/geocoding/v5/mapbox.places/${loc}.json?access_token=${mapboxAPIKey}`;
+    const coords = await axios.get(url);
+    coords.data && res.status(200).send({'latitude': coords.data.features[0].center[1], 'longitude': coords.data.features[0].center[0]});
```
wprater: To help readability, I'd suggest de-structuring the response like so:

```
if (coords && coords.data) {
  const [longitude, latitude] = coords.data.features[0].center;
  res.status(200).send({latitude, longitude});
} else { 
  // throw/fail 
}
```

rest-server/src/components/mapbox/mapboxRouter.js
```diff
@@ -0,0 +1,9 @@
+import express from 'express';
+import { fetchUserLocationCoordinates } from './mapboxControllers';
+
+const router = express.Router();
+
+router.route('/:_id')
```
wprater: Curious as to why this has it's own route vs. a endpoint from `users`?

rest-server/src/components/users/userControllers.js
```diff
@@ -0,0 +1,65 @@
...
+  try {
+    if (Object.keys(req.body) === 4) {
```
wprater: Using a magic number here is a bit fragile.  Consider leveraging your `userSchema` item instead.

rest-server/src/components/users/userControllers.js
```diff
@@ -0,0 +1,65 @@
...
+      const updated = await updateUserHelper(req.params, req.body);
+      res.status(201).send(updated);
+    } else {
```
wprater: This else block might deserve to be a `PATCH` route instead.

rest-server/src/components/users/userHelpers.js
```diff
@@ -0,0 +1,17 @@
+import { User } from '../../config/database/collections/userCollections';
+
+const addUserHelper = ({ name, dob, address, description }) => new User({
```
wprater, line 3: What is your reasoning for all the helper methods here which use the mongoose API over just using the API directly in the controllers?

rest-server/src/lib/log/index.js
```diff
@@ -0,0 +1,9 @@
+import * as chalk from 'chalk';
+
+export const success = (...log) => {
+  if (process.env.DEBUG === 'TRUE') console.log(chalk.default.white.bgGreen.bold(...log));
```
wprater: What happens when debug isn't set or does not evaluate to `TRUE`?

wprater (changes requested):
@pkwu great job! And thanks for your attention to details and commitment to get this up and documented.  I made some comments I'd like for you to review.

Look forward to hearing how your call with Codin goes today!

cc @codin-sf @aaron-sf

rest-server/src/config/express/index.js
wprater: @codin-sf  `OPTIONS` is required by the cors handler.

bin/buildEnv.js
pkwu: I have updated my entry file to check for the required ENV properties. Changes will be reflected in my next push.

bin/buildEnv.js
pkwu: I read from the creators of lodash that their each method performs faster than the native Array#forEach method. Importing the entire library is an oversight by me. I have updated my build file to import solely the each module.

docs/rest-server.md
pkwu: I will have swagger documentation available at 'localhost:3396/api-docs' in my next push.

package.json
pkwu: You are correct, they are certainly crucial modules necessary for the server to perform. I will fix this.

rest-server/bin/entry.js
pkwu: I was under the impression that by requiring babel-register I'd be enabling further files required by node which use es6 (i.e, import statement) to be transformed by Babel. I will look into this in further detail to better understand alternative approaches.

rest-server/bin/entry.js
pkwu: I have added an extra condition to check for the properties to exist in the env configuration object instead.

rest-server/src/components/mapbox/mapboxControllers.js
pkwu: That is a great suggestion! I didn't think to use array destructuring here. I will clean this up.

rest-server/bin/entry.js
wprater: Im not convinced you need to use Babel.  For which features are you transpiling for? `async / await` is available (for example) in `LTS`

rest-server/src/components/mapbox/mapboxRouter.js
pkwu: I treated the mapbox external API call as a separate route from the user route because it went beyond just accessing the user data from mongoDB. I figured since it's also making a request to the mapbox API, I'd leave it out of the basic CRUD function routes to avoid clutter. I can fix this.

wprater: Nothing you need to do here, I just wanted to understand your thinking.  Thank you.

rest-server/src/components/users/userControllers.js
pkwu: I will fix this.

rest-server/src/components/users/userControllers.js
pkwu: Ah, you are correct. When I was brainstorming how to handle the PUT route, I had to decide whether or not to maintain the current records that were not provided in the request. I opted to maintain the current records. However, on that note, I will fix this and separate the two http methods.

rest-server/src/components/users/userHelpers.js
pkwu: It is a habit of mine to abstract away the query logic, I can see how this style is more beneficial when writing SQL queries. However, if I were to directly call the mongoose API methods in the userController, I'd have a larger block of code. 

For example, updateUser in userController:

    const updated = await updateUserHelper(req.params, req.body);
    .
    .
    becomes
    .
    .
    const updated = await User.updateOne(req.params, {
      name: req.body.name, 
      dob: req.body.dob,
      address: req.body.address,
      description: req.body.description
    });

rest-server/src/lib/log/index.js
pkwu: This is a mistake on my part. I should have accounted for that. While the server runs, the server console logs stop displaying altogether if the env property does not evaluate to TRUE. I will fix this.

rest-server/src/components/users/userControllers.js
pkwu: You are correct, I got ahead of myself there and overthought the PUT handler. I have made the adjustments to separate PUT/PATCH and will push my changes soon.

rest-server/src/components/mapbox/mapboxControllers.js
pkwu: Great suggestion, I fiddled with the string to come up with my manual method. I have replaced it with JavaScript's global encodeURI function.

rest-server/test/user.js
pkwu: Awesome, I looked into the deep-eql plugin and will update my tests to use it accordingly.

rest-server/src/components/mapbox/mapboxControllers.js
pkwu: Cool! I've updated the Dockerfile to reflect that change. Would this still apply if it was an API key that could potentially be billed for? I ask this because I've read about not pushing up your keys to public repositories.

rest-server/bin/entry.js
pkwu: Currently if I remove my require statement for babel-register I'll run into errors with using import statements in node.

I've found this on stackoverflow as a preferred alternative to babel: 
https://nodejs.org/api/esm.html

I will try to enable the import statements through this method instead.

rest-server/src/components/mapbox/mapboxRouter.js
pkwu: Generally, would you consider it to be better practice keeping the mapbox API route as an endpoint on the users route because it is still related to the user model?

rest-server/bin/entry.js
pkwu: I checked out a separate branch on my forked repo to try to work around using babel. I've successfully managed to get the server + db up and running again using the .mjs extension. 

I'm still working on getting the tests running with mocha as it doesn't seem to play well with the .mjs files.

pkwu: Is using the native node esm .mjs extension (experimental) the best practice? 

I read about @std/esm and was able to get the server + db + tests + docker up and running using the esm package. 

I checked out a separate branch for this as well and have pushed my work to it. This can be viewed here: 
https://github.com/pkwu/node-backend-test/tree/esm

Is this the better approach for production apps?
I will merge the esm branch to my master branch to reflect the changes if approved.

PR #4, opened 2018-10-02
Superformula user backend

samples-afaq-tariq:
Documentation is in `docs.md`
Contribution Guidelines are in `contribution_guidelines.md`

Each item with a check mark below is something I've attempted to address in the code or docs.
 
#### Functionality

- [x] The API should follow typical RESTful API design pattern.
- [x] The data should be saved in the DB.
- [x] Provide proper API documentation.
- [x] Proper error handling should be used.

#### Bonus
- [x] Write clear documentation on how it has been designed and how to run the code.
- [x] Provide proper unit tests.
- [ ] Add a read only endpoint to fetch location information based off the user's address (use NASA or Mapbox APIs)
- [x] Use Docker containers.
- [x] Utilize Docker Compose.
- [ ] Setup a CircleCI config to build/test/deploy the service.
- [x] Write concise and clear commit messages.
- [ ] Leverage Terraform or other infrastructure management tooling to spin up needed resources.
- [ ] Providing an online demo is welcomed, but not required.

#### Advanced requirements

- [ ] Use hapi to build the core feature and use a different framework (such as Express or Loopback) to handle HTTP requests.
- [x] Provide a complete user auth (authentication/authorization/etc) strategy, such as OAuth.
- [x] Provide a complete error handling and logging (when/how/etc) strategy.
- [ ] Use a NoSQL DB and build a filter feature that can filter records with some of the attributes such as username. Do not use query languages such as MongoDB Query or Couchbase N1QL.


package.json
```diff
@@ -17,6 +17,7 @@
   },
   "homepage": "https://github.com/samples-afaq-tariq/node-backend-test#readme",
   "dependencies": {
-    "hapi": "^17.6.0"
+    "hapi": "^17.6.0",
+    "joi": "^13.7.0"
```
wprater: when adding dependencies, we like to include the lockfile or shrinkwrap, etc. so its easier to deploy and keep track of our dependencies.

src/users-plugin.js
```diff
@@ -3,26 +3,42 @@
...
+     * @return {Object: User} New user object
+     */
     server.route({
-      method:'GET',
+      method:'POST',
       path:'/',
+      options: {
+        validate: {
+          payload: {
+            name: joi.string().min(1).max(100).required()
```
wprater: should leverage the Joi user model setup in the previous commit.

package.json
```diff
@@ -18,6 +18,8 @@
   "homepage": "https://github.com/samples-afaq-tariq/node-backend-test#readme",
   "dependencies": {
     "hapi": "^17.6.0",
-    "joi": "^13.7.0"
+    "joi": "^13.7.0",
+    "lodash": "^4.17.11",
+    "uuid": "^3.3.2"
```
wprater: Prefer to add new dependencies (and lockfiles et al) as a discrete commit.

src/users-plugin.js
```diff
@@ -7,6 +7,12 @@ const joi                  = require('joi')
 const generateUUID         = require('uuid/v1')
 const userValidationSchema = require('./user-model')
 const pickProperties       = require('lodash/pick')
+const sendApiResponse = (type, responseValue) => {
+  return {
+    type: type,
```
wprater: Prefer to use shorthand property names on `type` and could rename `responseValue` to `item` or do you prefer the verboseness of the arg `responseValue`?

src/users-plugin.js
```diff
@@ -73,7 +78,7 @@ module.exports = {
     server.route({
       method:'DELETE',
       path:'/{userId}',
-      handler(request, h) {
+      handler(request, responseHandler) {
```
wprater: Any reason to keep around the `responseHandler` arg that's never used?

test/create-new-user.test.js
```diff
@@ -0,0 +1,43 @@
...
+  })
+
+  lab.test('should generate a system guid', async () => {
+
+    // these must match the route you want to test
+    const injectOptions = {
+      method: 'POST',
+      url: '/',
+      payload: {
+        name: "Beverly Benedict",
```
wprater, line 59: We will use prettier js and other linting tools to catch inconsistencies like these (quotes).  Have you worked with linting tools and prettier before?

docs.md
```diff
@@ -3,21 +3,39 @@
...
+    - indexes
+Testing
+Runtime
+  - Docker
+  - Logging
+    - log level guidance
+    - request logging vs. error logging
+    - log aggregation and insights strategy
+      - everything to stdout and aggregated by runtime process
+      - removes burden from app; can have a centralized logging library and collection process
```
wprater: Good idea here!

src/index.js
```diff
@@ -1,36 +1,52 @@
...
+    await usersApi.register(userServicePlugin, {
       routes: {
         prefix: '/api/v1/users'
       }
     })
 
-    await server.start()
-    
-    console.log(`superformula-users-api started on ${process.env.USERS_API_HOST}:${process.env.USERS_API_PORT}`)
+    await usersApi.start()
```
wprater, line 72: You can improve startup time by using the promise references for all the plugin registration vs. awaiting each one to resolve sequentially. 

```javascript
const plugin1Registration =  usersApi.register(plugin1)
const plugin2Registration =  usersApi.register(plugin2)
const plugin3Registration =  usersApi.register(plugin3)

await Promise.all([
  plugin3Registration, 
  plugin2Registration, 
  plugin1Registration
])
  .catch(() => /* figure out which plugin failed to register */ )

await usersApi.start()
```

docs.md
```diff
@@ -3,11 +3,23 @@
...
+Access
+  Authentication
+  Authorization
 Hapi Server
   - RESTful
+    - no PATCH or LIST. simpler behaviors
+      - has danger of being misused if there are single-property consuming downstream services for PUT.
   - Error handling
+  - infrastructural app integrity
+    - cutoff long responses with 50x
```
wprater: Can you expound upon your thinking here?

docs.md
```diff
@@ -3,17 +3,134 @@
...
+##### UPDATE USER
+PUT /api/v1/users/{userId}
+
+Response Type: Single Entity (User)
+
+`userId` must be the `id` property of an existing user.
+
+Read-only properties such as `id`, `createdAt`, and `updatedAt` cannot be modified and cannot be provided in the request body.
+
+This endpoint allows for replacing of the existing user record with the one provided. This means that any ommitted properties will be deleted from the record.
```
wprater: Agree, there is alway a danger here.  Which is why it's always nice to have a PATCH :)

src/route-handlers/create-new-user.js
```diff
@@ -8,18 +9,37 @@ const generateUUID = require('uuid/v1')
...
     createdAt: new Date().toISOString(),
   }
 
   // assure updatedAt is exactly the same as createdAt at this point.
   systemGeneratedProperties.updatedAt = systemGeneratedProperties.createdAt
-
   const newUser = Object.assign(request.payload, systemGeneratedProperties)
+  
+  // rename id to _id for mongoDB's sake
+  newUser._id = newUser.id
```
wprater, line 21: Will likely wish to abstract this later if/when we change data stores.

src/route-handlers/update-user-by-id.js
```diff
@@ -0,0 +1,38 @@
...
+module.exports = async function updateUserById(request, responseHandler) {
+  try {
+    const userCollection = request.mongo.db.collection('users')
+    const fetchedUser    = await userCollection.findOne({ _id: request.params.userId })
+
+    if (!fetchedUser) {
+      return boom.notFound('User with provided id was not found.')
+    } else {
+
+      const updatedUserRecord = Object.assign(request.payload, {
```
wprater: May be dangerous to augment the `request.payload` here in case we need to log this in it's originality?  In this case we'd `const updatedUserRecord = Object.assign({}, request.payload, {...})`

src/users-plugin.js
```diff
@@ -3,61 +3,89 @@
...
+        handler: listUsersHandler
+      },
+      {
+        path:'/{userId}',
+        method:'GET',
+        options: {
+          tags: ['api', 'user'],
+          validate: {
+            params: {
+              userId: joi.string().guid()
```
wprater, line 55: Should this validation be pulled from the the user schema object instead?

src/route-handlers/delete-user-by-id.js
```diff
@@ -9,8 +9,22 @@ const boom = require('boom')
...
-    await request.mongo.db.collection('users').deleteOne({ _id: request.params.userId })
-    return {}
+    const userRecord = await request.mongo.db
+      .collection('users')
+      .findOne({ _id: request.params.userId, archived: false })
+
+    if (!userRecord) {
+      return boom.notFound()
+    } else {
+      userRecord.archived = true
```
wprater, line 18: Do we need to update any tests for this change in behavior?

src/route-handlers/update-user-by-id.js
```diff
@@ -16,14 +15,13 @@ module.exports = async function updateUserById(request, responseHandler) {
       return boom.notFound('User with provided id was not found.')
     } else {
 
-      const updatedUserRecord = Object.assign(request.payload, {
-        createdAt: fetchedUser.createdAt,
+      const updatedUserRecord = Object.assign(fetchedUser, request.payload, {
```
wprater, line 17: 👍

src/index.js
```diff
@@ -57,6 +57,18 @@ async function startApi() {
...
+    usersApi.method('getUserById', userId => mongoUserCollection.findOne({ _id: userId, archived: false }))
+    usersApi.method('insertUser', newUser => mongoUserCollection.insertOne(newUser))
+    usersApi.method('updateUser', (userId, user) => mongoUserCollection.updateOne({ _id: userId }, user))
+    usersApi.method('getAllUsers', (limit, skip) => {
+      return mongoUserCollection.find({ archived: false }, { name: true, id: true, updatedAt: true })
+        .sort({ updatedAt: -1 })
+        .limit(limit)
+        .skip(skip)
+        .toArray()
+    })
```
wprater, line 70: 👍

contribution-guidelines.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. Make code change
+2. Make calls to the API on port 8081
+3. Commit your changes
+
+## Remote Build
+
+When building the project on a remote build process such as CircleCI, AWS Codebuild, or Jenkins, use the following process:
+
+1. Run `npm install --production`
```
wprater, line 30: 👍

wprater (commented):
@samples-afaq-tariq thanks for your submission.  Left some comments and small improvements.  Wonderful job with documentation, process, and best practices.  Like the part about logging.  No changes are needed, but would like to hear comments to some of my changes or questions. 

Thank you!

ps - I did my comments commit by commit, and a couple of them look to have been corrected in future commits.

package.json
samples-afaq-tariq: I committed `package-lock.json` from early on under the impression that it would be the lock-file used by all npm install operations from that point forward. Is that not the case? Or are you referring to something else I'm missing here?

src/index.js
samples-afaq-tariq: Makes sense. Good to know. Hadn't thought of that.

src/route-handlers/create-new-user.js
samples-afaq-tariq: Fair point. This is a bit of an oversight on my part. I tried to abstract out other aspects of mongoDB from request handlers into server methods. This one could be another one for general purpose schema transformation/translation.

src/route-handlers/delete-user-by-id.js
samples-afaq-tariq: Not sure what you mean. There is a `delete-user.test.js` file that has a test case that accounts for `archived` property being set when delete is called.

Was there something else here that you're suggesting be tested?

src/users-plugin.js
samples-afaq-tariq: yes, it should. I should have written it as `userId: userValidationSchema.id` instead.

Also, I walked myself into a regrettable pattern of adding `.default(...)` to some of the user schema properties. While it still would "work" in this scenario, it is an awkward pattern to get default values from a model's schema that is only really used for validation. I would undo that decision next.

test/create-new-user.test.js
samples-afaq-tariq: Yes. All of my current work projects install an internal package that contains our base linting rules, code formatting rules, and helper webpack/rollup configs. They run on local build, on git hook, and cause remote builds to fail if linting rules don't pass. 

The convenience of them being part of our base repo intialization template has probably spoiled me to the point that I didn't even think about installing them here. Clearly, that should have happened. Its very helpful to have standard code formatting, and code linting.

docs.md
samples-afaq-tariq: The idea I was going for was a hard-defined upper limit on duration / response times of all HTTP calls. They would be terminated at the infrastructure layer (load balancer, edge gateway, etc.).

In larger microservices ecosystems, I've seen this be an effective means to:
- proactively deter processes from introducing unacceptable response times sporadically
- force engineers to account for resource-intensive operations through more effective means such as introducing eventual consistency or more compute/memory resources
- allow you to have a stricter alignment to percentile request time targets in SLAs
- surface long-running-responses as true failures (50x) rather than just an outlier or skewing metric.

It definitely comes with some caveats that then need to be thought about. For example, cutting off a request doesn't mean the request process was killed. And subsequent calls (retry) by the client raise the question of idempotent nature of that action.

src/route-handlers/update-user-by-id.js
samples-afaq-tariq: Yea, that's a very good point. It shouldn't be written that way at all. I would correct that everywhere I used `Object.assign`.

This can also introduce bugs as its easy to overlook that request.payload is being modified. Better to lean towards functional-paradigm of not changing state.

src/users-plugin.js
samples-afaq-tariq: My only preference is clarity. I've never found verboseness helpful by itself.

`item` seems clear enough too, considering the function's and parameter's  original purpose. 

agree on `type` vs `type: type`.

src/users-plugin.js
samples-afaq-tariq: No good reason. I was originally expecting to use `responseHandler` extensively as I got accustomed to `hapi` patterns. But that didn't pan out and I should have removed all references.

I'm also realizing that had I added a linter, it would have picked up on `responseHandler` as an unused var as well.

src/route-handlers/delete-user-by-id.js
wprater: This was all, unfortunately I had written this comment on a commit before Id seen the updated test I think!

test/create-new-user.test.js
wprater: Yes, it's very helpful; so much so, I can see how it could be forgotten.  Honestly, we should be to a point where we dont need to think about formatting on input, as it should be authoritatively stored formatted as such.  Go started this trend with gmfmt and Im happy to see other following suit!

wprater:
Thanks again for your submission.  Will be in touch shortly.

PR #6, opened 2019-02-20
demo

ghost:


README
```diff
@@ -0,0 +1,44 @@
...
+Some highlights:
+    - the user model tests against a local dynamo instance
+    - each test starts with a clean db to help isolate each test case
+    - a coverage report is available using nyc
+    - some ideas for how metrics and logging might work are included in comments
+    - a basic circle ci config is present but not tested, if not spot on, think it's close.
+
+Code quirks?
+    - I know it's unpopular, but I like to use let instead of const..
+    - It's less characters and let's me keep things dynamic (like the language).
```
tmarciszewski, line 35: I'd like to hear more about it.

app.spec.js
```diff
@@ -0,0 +1,401 @@
...
+		assert(iso8601Pattern.test(inserted.createdAt));
+		assert(iso8601Pattern.test(inserted.updatedAt));
+
+		let fetched = User.fetch(inserted.id);
+		for (let k of Object.keys(fetched)) {
+			assert(fetched[k] === inserted[k]);
+		}
+	});
+
+	it('rejects invalid data on user post', async function () {
```
tmarciszewski, line 167: I think splitting this huge test into several smaller ones is beneficial.

app.js
```diff
@@ -0,0 +1,198 @@
...
+		return;
+	}
+
+	let duration = Date.now() - start;
+	metrics.record('lat_auth_authorized', duration);
+	next();
+});
+
+app.get('/v1/users', async (req, res) => {
+	let start = Date.now();
```
tmarciszewski, line 62: I don't really like the boilerplate-to-actual-functionality ratio. While I understand in some cases it's important to measure in-handler metrics you could get away with just writing simple middleware which would do that for you automatically and without cluttering the code.

misiek-sf:
@sf-r can you please point me to the part that you are the most (and least) proud of?

app.js
codin-mgm: @sf-r if you were going to approach this, how would you do it?

ghost: agreed, you're right @tmarciszewski 

think we can get around this by utilizing `next()` in our existing mws and sandwich them with two `use` mws that make use of res.locals 👍 

thoughts?

README
ghost: it's mostly a combination of muscle memory, ocd and allowing for variable reuse to avoid having to come up with more variable names... it's just a quirk, I know it's sacrilege ;)

ghost: PS: that is an awesome profile pic.. like for real.

app.spec.js
ghost: yeah, that's fair.. I guess my mind grouped them together initially because of the similarity but splitting out def doesn't hurt here, and pretty simple to do.

ghost:
hey @misiek-sf , sorry I missed this request earlier. That's a great question.  Rather than pointing to anything specific, I think as a first pass, I'm feeling decent about the level of code coverage and the testing strategy overall.  The coverage is pretty high and I think the way the user model is structured is handy.  I also feel like there's a good amount of vanilla, which hopefully resonates a bit with others?  Less digging to understand how things are working, more empowerment perhaps, maybe also something that may help with code shelf life?  (if anyone agrees, it wasn't something that required a lot of effort..)

I guess I was also thinking that people sometimes shirk on the testing? Which I thought might be a good thing to demonstrate.  That said, I probably rushed on some of the decisions around the api interface itself.. some more checks like, does the record already exist on create, are we trying to delete a non existent record?  A consistent payload format for all responses.

I guess that feels like the type of details you'd start filling in after some of the broader strokes laid out here?  That and functionality for batch operations might be a nice touch too. 

@tmarciszewski gave me a good poke to give some thought to removing some boiler plate from the middleware... curious to hear what your takeaways are as well... hope this kind of answers the q.

app.js
tmarciszewski: Yeah. That's probably what I'd do. You have full control over order of middleware execution so it's quite easy to move the most repeatable code to two middlewares which would execute at the start and at the end of request.

models/user.js
```diff
@@ -0,0 +1,153 @@
...
+		return errors.length ? errors : null;
+	}
+
+	/**
+	 * @param {User} user
+	 * @return {User|Error} inserted
+	 */
+	static async insert(user) {
+		if (!user) {
+			return new Error('missing user');
```
tmarciszewski, line 88: I'd advise against returning errors. It should be thrown because in this case you need to always check for type of returned value instead of using more natural try/catch mechanisms.

models/user.js
```diff
@@ -0,0 +1,153 @@
...
+		if (typeof o === 'string') {
+			o = JSON.parse(o);
+		}
+		for (let k of Object.keys(this)) {
+			if (o[k] === undefined) {
+				this[k] = undefined;
+				continue;
+			}
+
+			this[k] = JSON.parse(JSON.stringify(o[k]));
```
tmarciszewski, line 37: Is there a reason for all this cloning? We are dealing with simple values here, so references are not an issue.

models/user.js
```diff
@@ -0,0 +1,153 @@
...
+ * @class User
+ * @property {string} id
+ * @property {string} name
+ * @property {string} dob
+ * @property {string} address
+ * @property {string} description
+ * @property {string} createdAt
+ * @property {string} updatedAt
+ */
+class User {
```
tmarciszewski, line 14: While fine for simple demo purpose I'd suggest separating concerns of such classes. Data model is one thing. Validation is another. Operations on the database should also be separated. This has many advantages including much easier code reuse and plain old "keeping it simple".

tmarciszewski:
I've added few more comments. Just general advice in most cases.

One more thing I'd like you to explain is why have you opted to roll out your own tools instead of using something ready to use. Like axios/request instead of your own HttpClient. Or something which would streamline validation instead of "manual" checking.

models/user.js
ghost: totally, this is actually more of a conversation piece because I've seen some people talk about adopting this type of convention based on experiencing something similar with golang and a desire to eliminate some of the baggage that comes with try/catch.

Try/catch is def idiomatic w js and something I'm totally cool with using when working with a team.  For personal projects I might do something like this because I like how it feels, but definitely wouldn't try to push it on others.

models/user.js
ghost: Yeah I hear that too.  Are you thinking Object.assign() for a shallow copy? I just err on the side of this strategy to avoid situations where something changes and now I've got shared references to deeply nested objects.

Were you thinking Object.assign() here?

models/user.js
ghost: yeah, I think that's fair.  I can totally see it being set up the way you're suggesting and working nicely.  I think I started the work with a  separate layer for the db, with a couple of small tweaks I think we can flesh it out just a bit more so the other methods related to DB ops aren't needed here.  Then we can stub out db layer methods rather than methods from here in the tests.  

Open to other suggestions as well.

models/user.js
tmarciszewski: In this case simple assignment would work perfectly fine. Maybe if we really add nested objects I’d add special case for it and clone only what’s necesary. Object.assign does shallow copy so it’s probably not that usefull. Stringify/parse pattern, while horribly ugly, is still the most performant way of doing deep clone of simple objects :-)

ghost:
Totally, for the http client specifically.. I haven't done it in this demo but in the past I like to add some basic functionality that makes it easy for me stub responses based on type of requests, and order of request and so on.  The code for it is minimal and has served my needs okay..

That said, I can totally use request.. but then I also need to find the mocking library which maybe was mock-request on my first google search? But that got deprecated, so next I'm looking at nock.. and this blog post is telling me if I want to use nock I really should be using axios... so now I'm figuring out the interface for axios and nock, and next month everyone says stop using whatever you're using because everyone has changed their minds (but I'm still just making http requests).  I guess that's why I sometimes err on the side of caution when I don't think there's a heavy lift to make it work with stuff from core.  For validation it's similar for me.. I can control all the dials and do it in what feels like a relatively simple fashion.  But I'm not knocking any of that stuff and I'm not necessarily advocating that anyone should do anything differently!  

I guess the demo here is part demonstration and part self-expression, though the latter doesn't necessarily need to reflect how I'd approach things when coordinating with a team..  If anything, maybe it mainly creates an opportunity to shoot the sh*t about code ;)

wprater:
I thought Nock just monkey patched stdlib http, so you could use it with
any other request abstraction lib?  That being said, the last time I used
Nock was with Axios :)  And the new kid on the block seems to be got
https://github.com/sindresorhus/got

ghost:
😆there you have it :)

tmarciszewski:
> And the new kid on the block seems to be got

I think for many developers `request` is still the go-to tool when it comes to making requests. It probably got less popular in the last years because the interface is very strange (too many and not very consistent way of defining request options). Also its refusal to endorse promises forcing everyone to use another library/wrapper or rolling out your own boilerplate. Request works fine with nock but there's some cleanup you may need to do after each test to have it working correctly.

ghost:
yeah, re: popularity I think that's probably right @tmarciszewski.  
quick look at npm: 13,172,463 weekly downloads!

ghost:
check this out... looks like he decided he wanted to rewrite it in 2017
https://github.com/mikeal/r2

ghost:
bummer.. gl.

PR #7, opened 2019-03-05
Superformula Coding Exercise

cudom:
Here is my implementation of a RESTful API that can create/read/update/delete user data from a persistence store. With permission from Codin, I took some creative liberties with the requirements to possibly leverage my work for use in my golf performance training business. Having only worked for about a year with Node.js on AWS Lambda without a Node.js framework, this was a great opportunity for me to learn Express or hapi.

My approach to completing this coding exercise was to first became familiar with the new technology and experiment with running examples. I have always believed it is important to stand up a lightweight but working end-to-end solution first. I did not want clutter the commit history with the learning and experimenting phase. Therefore, my initial commit included a basic implementation that works end-to-end for a few of the endpoints with the remaining ones stubbed out. It also included the project framework for unit and integration testing with coverage, standardized code style, and linting. I then incrementally implemented the remaining functionality one endpoint at a time along with its associated tests.

Documentation on how the solution was designed and how to run the code is in the README.md. Please note that the documentation and testing is much lighter than I would like it to be. My original plan was to complete all the “Bonus” and “Advanced requirements” tasks prior to submission. There are still a lot of little things to do like pulling the database URL out into an environment variable and making a config file based on the deployment environment.

Given my time constraints due to working full-time and family obligations, I am submitting my work for review now. Regardless of the outcome of this interview process, I plan to continue to improve the documentation and add new features to this project. I will also complete all the “Bonus” and “Advanced requirements” tasks for my own benefit and learning. 

That said, I am looking forward to the code review and receiving feedback on my work.

### DIFF
--- a/bin/seed-golfathlete-db.js
+++ b/bin/seed-golfathlete-db.js
@@ -45,7 +45,7 @@ function createAthlete (lastName, firstName, ghinNumber, cachedGhinIndex, cached
 function createAthletes (cb) {
   async.series([
     function (callback) {
-      createAthlete('Udom', 'Charlie', '5192665', '8.3', '2019-03-02', '106.4', '1973-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'TPI Level 2 Coach', callback)
+      createAthlete('Udom', 'Charlie', '5192665', '8.3', '2019-03-02', '106.4', '1973-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'TPI Level 2 Certified Coach', callback)
     },
     function (callback) {
       createAthlete('Udom', 'Alexa', '0146376', '+2.0', '2019-03-02', '97.1', '2002-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'Future NCAA D1 Golfer', callback)
--- a/bin/seed-golfathlete-db.js
+++ b/bin/seed-golfathlete-db.js
@@ -3,9 +3,10 @@
 'use strict'
 
 const async = require('async')
-const Athlete = require('../src/models/athlete')
-
 const mongoose = require('mongoose')
+const Athlete = require('../src/models/athlete')
+const athleteData = require('./athlete-data')
+
 mongoose.set('useCreateIndex', true)
 mongoose.connect('mongodb://localhost:27017/golfAthleteTest', { useNewUrlParser: true })
 mongoose.Promise = global.Promise
@@ -14,44 +15,35 @@ db.on('error', console.error.bind(console, 'MongoDB connection error:'))
 
 const athletes = []
 
-function createAthlete (lastName, firstName, ghinNumber, cachedGhinIndex, cachedGhinIndexDate, driverClubHeadSpeed, dob, address, description, cb) {
+function createAthlete (athleteData, cb) {
   const now = Date.now()
 
   const athlete = new Athlete({
-    lastName: lastName,
-    firstName: firstName,
-    ghinNumber: ghinNumber,
-    cachedGhinIndex: cachedGhinIndex,
-    cachedGhinIndexDate: cachedGhinIndexDate,
-    driverClubHeadSpeed: driverClubHeadSpeed,
-    dob: dob,
-    address: address,
-    description: description,
+    ...athleteData,
     createdAt: now,
     updatedAt: now
   })
-
-  athlete.save(function (err) {
-    if (err) {
+  athlete.save()
+    .then(newAthlete => {
+      console.log('New Athlete: ' + newAthlete)
+      athletes.push(newAthlete)
+      cb(null, newAthlete)
+    })
+    .catch(err => {
       cb(err, null)
-      return
-    }
-    console.log('New Athlete: ' + athlete)
-    athletes.push(athlete)
-    cb(null, athlete)
-  })
+    })
 }
 
 function createAthletes (cb) {
   async.series([
     function (callback) {
-      createAthlete('Udom', 'Charlie', '5192665', '8.3', '2019-03-02', '106.4', '1973-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'TPI Level 2 Certified Coach', callback)
+      createAthlete(athleteData.charlie, callback)
     },
     function (callback) {
-      createAthlete('Udom', 'Alexa', '0146376', '+2.0', '2019-03-02', '97.1', '2002-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'Future NCAA D1 Golfer', callback)
+      createAthlete(athleteData.alexa, callback)
     },
     function (callback) {
-      createAthlete('Udom', 'Momo', '1239208', '6.9', '2019-03-02', '85.3', '2004-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'Freshman on the Varsity Golf Team', callback)
+      createAthlete(athleteData.momo, callback)
     }
   ],
   cb)
@@ -59,6 +51,7 @@ function createAthletes (cb) {
 
 async.series([
   createAthletes
+  // add create methods for SuperSpeed and TPI data
 ],
 function (err, results) {
   if (err) {
--- a/bin/seed-golfathlete-db.js
+++ b/bin/seed-golfathlete-db.js
@@ -16,8 +16,6 @@ const athletes = []
 function createAthlete (lastName, firstName, ghinNumber, cachedGhinIndex, cachedGhinIndexDate, driverClubHeadSpeed, dob, address, description, cb) {
   const now = Date.now()
 
-  // todo joi validate
-
   const athlete = new Athlete({
     lastName: lastName,
     firstName: firstName,
@@ -52,7 +50,7 @@ function createAthletes (cb) {
       createAthlete('Udom', 'Alexa', '0146376', '+2.0', '2019-03-02', '97.1', '2002-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'Future NCAA D1 Golfer', callback)
     },
     function (callback) {
-      createAthlete('Udom', 'Momo', '1239208 ', '6.9', '2019-03-02', '85.3', '2004-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'Fresh on the Varsity Golf Team', callback)
+      createAthlete('Udom', 'Momo', '1239208', '6.9', '2019-03-02', '85.3', '2004-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'Fresh on the Varsity Golf Team', callback)
     }
   ],
   cb)
--- a/bin/seed-golfathlete-db.js
+++ b/bin/seed-golfathlete-db.js
@@ -6,7 +6,8 @@ const async = require('async')
 const Athlete = require('../src/models/athlete')
 
 const mongoose = require('mongoose')
-mongoose.connect('mongodb://localhost:27017/golfAthlete', { useNewUrlParser: true })
+mongoose.set('useCreateIndex', true)
+mongoose.connect('mongodb://localhost:27017/golfAthleteTest', { useNewUrlParser: true })
 mongoose.Promise = global.Promise
 const db = mongoose.connection
 db.on('error', console.error.bind(console, 'MongoDB connection error:'))
@@ -50,7 +51,7 @@ function createAthletes (cb) {
       createAthlete('Udom', 'Alexa', '0146376', '+2.0', '2019-03-02', '97.1', '2002-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'Future NCAA D1 Golfer', callback)
     },
     function (callback) {
-      createAthlete('Udom', 'Momo', '1239208', '6.9', '2019-03-02', '85.3', '2004-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'Fresh on the Varsity Golf Team', callback)
+      createAthlete('Udom', 'Momo', '1239208', '6.9', '2019-03-02', '85.3', '2004-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'Freshman on the Varsity Golf Team', callback)
     }
   ],
   cb)
--- a/bin/seed-golfathlete-db.js
+++ b/bin/seed-golfathlete-db.js
@@ -4,14 +4,16 @@
 
 const async = require('async')
 const mongoose = require('mongoose')
+const config = require('../src/config')
 const Athlete = require('../src/models/athlete')
 const athleteData = require('./athlete-data')
+const logger = require('../src/logger')
 
 mongoose.set('useCreateIndex', true)
-mongoose.connect('mongodb://localhost:27017/golfAthleteTest', { useNewUrlParser: true })
+mongoose.connect(`${config.db.url}${config.db.name}`, { useNewUrlParser: true })
 mongoose.Promise = global.Promise
 const db = mongoose.connection
-db.on('error', console.error.bind(console, 'MongoDB connection error:'))
+db.on('error', logger.error.bind(logger, 'MongoDB connection error:'))
 
 const athletes = []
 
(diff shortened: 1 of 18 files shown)


bin/seed-golfathlete-db.js
```diff
@@ -0,0 +1,70 @@
...
+const mongoose = require('mongoose')
+mongoose.set('useCreateIndex', true)
+mongoose.connect('mongodb://localhost:27017/golfAthleteTest', { useNewUrlParser: true })
+mongoose.Promise = global.Promise
+const db = mongoose.connection
+db.on('error', console.error.bind(console, 'MongoDB connection error:'))
+
+const athletes = []
+
+function createAthlete (lastName, firstName, ghinNumber, cachedGhinIndex, cachedGhinIndexDate, driverClubHeadSpeed, dob, address, description, cb) {
```
tomasz-sf: Given JS language features (or lack of those features :P) it's usually not a good idea to have functions with more than 2-3 parameters. It's very easy to make mistakes and it's hard to refactor. In those cases it's better idea to provide all the required data as single object `{ lastName, firstName, ghinNumber, cachedGhinIndex, cachedGhinIndexDate, driverClubHeadSpeed }`

bin/seed-golfathlete-db.js
```diff
@@ -0,0 +1,70 @@
...
+  const now = Date.now()
+
+  const athlete = new Athlete({
```
tomasz-sf, line 23: It's a good idea to use shorthand notation making it
```
const athlete = new Athlete({
    lastName,
    firstName,
    ghinNumber,
    cachedGhinIndex,
    cachedGhinIndexDate,
    driverClubHeadSpeed,
    dob,
    address,
    description,
    createdAt: now,
    updatedAt: now
  })
```

Or even, if you used my suggestion from previous comment, you may shorten it to:
```
function createAthlete (athleteData, cb) {
  const now = Date.now()

  const athlete = new Athlete({
    ...athleteData,
    createdAt: now,
    updatedAt: now
  })

  athlete.save(function (err) {
    if (err) {
      cb(err, null)
      return
    }
    console.log('New Athlete: ' + athlete)
    athletes.push(athlete)
    cb(null, athlete)
  })
}
```

bin/seed-golfathlete-db.js
```diff
@@ -0,0 +1,70 @@
...
+    dob: dob,
+    address: address,
+    description: description,
+    createdAt: now,
+    updatedAt: now
+  })
+
+  athlete.save(function (err) {
+    if (err) {
+      cb(err, null)
```
tomasz-sf, line 35: You may shorten it to `return cb(err, null)`

bin/seed-golfathlete-db.js
misiek-sf, line 23: is there a reason not to use object shorthand?

src/controllers/athlete-controller.js
```diff
@@ -0,0 +1,94 @@
...
+  // todo index ghinNumber to avoid collection scan
+  Athlete.findOne({ ghinNumber: req.body.ghinNumber })
+    .exec((err, athlete) => {
+      if (err) {
+        return next(err)
+      } else if (athlete && athlete.ghinNumber && athlete.ghinNumber === req.body.ghinNumber) {
+        return next(boom.badRequest('Athlete with this GHIN number already exists'))
+      }
+      const now = Date.now()
+      const newAthlete = new Athlete(Object.assign({ _id: uuid(), createdAt: now, updatedAt: now }, req.body))
```
tomasz-sf: Is there a reason for using uuids instead of MongoDB ids?

src/models/athlete.js
```diff
@@ -0,0 +1,41 @@
+'use strict'
+
+const mongoose = require('mongoose')
+const joigoose = require('joigoose')(mongoose)
+const joi = require('joi')
+const uuid = require('uuid/v1')
```
misiek-sf, line 6: afair mongo uses `v4`

bin/seed-golfathlete-db.js
```diff
@@ -0,0 +1,70 @@
...
+      return
+    }
+    console.log('New Athlete: ' + athlete)
```
misiek-sf: maybe use `debug` if u already have it 😉

src/controllers/athlete-controller.js
```diff
@@ -0,0 +1,94 @@
...
+          return next(err)
+        } else {
+          res.status(201).json(util.fromMongoDb(result))
+        }
+      })
+    })
+}
+
+exports.updateAthlete = (req, res, next) => {
+  Athlete.findById(req.params.id)
```
tomasz-sf: You've added Promises to mongoose before (`mongoose.Promise = global.Promise`) so it would be a good idea to use them. Code would be much simpler.

bin/seed-golfathlete-db.js
```diff
@@ -0,0 +1,70 @@
...
+      createAthlete('Udom', 'Alexa', '0146376', '+2.0', '2019-03-02', '97.1', '2002-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'Future NCAA D1 Golfer', callback)
+    },
+    function (callback) {
+      createAthlete('Udom', 'Momo', '1239208', '6.9', '2019-03-02', '85.3', '2004-01-01', '4235 NW 192nd Ave, Portland, OR 97229', 'Freshman on the Varsity Golf Team', callback)
+    }
+  ],
+  cb)
+}
+
+async.series([
```
misiek-sf: why this is `series` invoking `series`

bin/seed-golfathlete-db.js
misiek-sf: maybe all `console.log` should/would be replaced by `debug`

test/app.spec.js
```diff
@@ -0,0 +1,216 @@
...
+        .end((err, res) => {
+          id = res.body.id
+          createdAt = res.body.createdAt
+          expect(res.body.lastName).to.equal('McTesterson')
+          done()
+        })
+    })
+  })
+
+  describe('POST /api/v1/athletes', () => {
```
tomasz-sf: `describe` is used to group tests with similar topic but you have several describes for `POST /api/v1/athletes`

package.json
```diff
@@ -0,0 +1,48 @@
...
+    "test": "npm run stylint && npm run test:coverage",
+    "test:coverage": "nyc --check-coverage --lines 95 --functions 95 --branches 95 npm run test:all",
+    "test:all": "mocha ./test/**/*.spec.js --timeout 5000 --exit"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/cudom/node-backend-test"
+  },
+  "dependencies": {
+    "acorn": "^6.1.1",
```
misiek-sf, line 22: why u need `acorn`?

test/app.spec.js
```diff
@@ -0,0 +1,216 @@
...
+        .expect('Content-Type', /json/)
+        .expect(404, done)
+    })
+  })
+
+  describe('PUT /api/v1/athletes/:id with last name missing', () => {
+    it('should return 400', done => {
+      request(app)
+        .put(`/api/v1/athletes/${id}`)
+        .send(Object.assign({}, athlete, { id: id, lastName: undefined }))
```
tomasz-sf: You can use spread operator here:
```.send({ ...athlete, id, lastName: undefined })```

**Review by tomasz-sf — COMMENTED** [review]
My general feeling is that you should invest some time in understanding and using Promises. Most of the code would be much more concise and readable if you replaced callbacks with Promises/async-await.

File: src/controllers/athlete-controller.js [hunk 12]
```diff
@@ -0,0 +1,94 @@
...
+    .exec((err, athlete) => {
+      if (err) {
+        return next(err)
+      } else if (athlete == null) {
+        return next(boom.notFound())
+      }
+      // do not allow the createdAt value to updated and set a the current date/time on the updatedAt
+      let updatedAthlete = new Athlete(Object.assign(util.toMongoDb(req.body), { _id: req.params.id, createdAt: athlete.createdAt, updatedAt: Date.now() }))
```
tomasz-sf: mongoose will update only the fields which were sent so you may just `delete req.body.createdAt`

It's also a good idea to not update those fields manually. Mongoose will handle it for you

src/controllers/athlete-controller.js
```diff
@@ -0,0 +1,94 @@
...
+    })
+}
+
+// implement using jsonpath; will be useful once the object graph becomes more complex
+exports.patchAthlete = (req, res, next) => {
+  res.send('NOT IMPLEMENTED')
+}
+
+exports.deleteAthlete = (req, res, next) => {
+  Athlete.findById(req.params.id)
```
misiek-sf: why you are doing two requests to DB instead of one?

misiek-sf (commented):
@cudom I've started to do the review but it seems that i first need some question/doubds to be answered...

* is there a reason why u are not using promisses but callbacks and `async`?
* what is reasoning behind the `_id`/`id` conversion? (why not to do `id` in the schema on the fristplace)
* can u please write something about why u are structuring your test that way (describe -> it, describe ->it)
* why not use spred operator instead of `Object.assign`

bin/seed-golfathlete-db.js
cudom: Thanks for the feedback. What you said makes sense to me and it will make the function signature cleaner. I will make that change.

bin/seed-golfathlete-db.js
cudom: I like the above code. Much more concise. Will change to the shortest version.

src/controllers/athlete-controller.js
cudom: I have always used uuids since I believe they are a generally accepted standard for unique ids. Also, I specifically choose v1 for uniqueness vs v4 for randomness since I am not concerned about the possibility of the loss of anonymity. In reality, I think either would be fine. I need to look more into MongoDB ids to see if they would have been acceptable.

test/app.spec.js
cudom: Thanks for this tip. Much more concise. Will make these changes as well.

src/controllers/athlete-controller.js
cudom: True. Deleting req.body.createdAt would have achieved the same goal. Noted that Mongoose will handle those timestamp fields automatically. Will clean that up.

bin/seed-golfathlete-db.js
cudom: 1. My preference is actually to use async/await whenever possible. Since Express, Mongoose, MongoDB is new technology for me, I followed the some examples as-is and have not gone back to update the code yet. I do feel it's important to have consistency in how Promises are used and do like the improved readability of async/await.
2. _id/id? Oversight on my part. I actually plan to pull out the model into its own file and will change _id to be id. This way the model would be more portable since it wouldn't have any MongoDB-ism in it.

cudom: 3. Regarding the structuring of the test, I was focused on getting the tests in place and working. I agree I need to group the tests more logically by describe -> it, it, it... etc. so they are easier to follow.
4. Tomasz pointed out using the spred operator also. I was not familiar with it. However, now that I am, I will definitely use it because I like the conciseness of it.

src/controllers/athlete-controller.js
tomasz-sf: This may make some sense. I don't remember the exact thing `deleteOne` returns. If it returns number of deleted documents you may skip the findById part and return 404 if it returns 0. 

In real life situations, you usually need to get the document before deleting it to check permissions, dependencies etc.

bin/seed-golfathlete-db.js
cudom: That was left as a placeholder. I plan on adding more collections related to training data to this seed file. Should have added a comment there to avoid confusion.

package.json
cudom: When I installed one of the other dependencies, I got an error about acorn also being required. I installed to move forward in the project and haven't had a chance to look further into it yet. I do think that unnecessary packages should not be included.

src/controllers/athlete-controller.js
cudom: I had some issues getting findByIdAndUpdate to work and ended up doing 2 calls here as well. I just followed the same pattern (not a good pattern) for delete since I knew it worked. I will circle back around and try to get the single call for both working.

bin/seed-golfathlete-db.js
cudom: Thanks.

src/controllers/athlete-controller.js
cudom: Agree. Will make the change. In my current project at work, my preference is to use async/wait when possible because the code is much easier to follow as you stated.

test/app.spec.js
cudom: That makes sense. Thanks. I was following a Mocha example and didn't realize I could do describe->it, it, it, etc. until you guys pointed it out. Will do the grouping to make the tests easier to follow.

cudom:
Thanks to both of you for your feedback and your time. Learned some new things that will improve my coding. I will make the suggested changes and commit them.

bin/seed-golfathlete-db.js
cudom: Yes. Just a habit. I was planning to remove console.log lines and missed doing that. For that matter, I still need to implement better logging throughout the code. Debug where it is needed and error logs as necessary.

src/models/athlete.js
cudom: I looked it up and you are correct.  I specifically choose v1 for uniqueness vs v4 for randomness since I am not concerned about the possibility of the loss of anonymity. In reality, I think either would be fine.

src/controllers/athlete-controller.js
cudom: That's a good point Tomasz. In my case, I may not want to allow deletion if there is training data that associated with the athlete. I did verify that findByIdAndDelete will return a 404 if it cannot find the doc.

bin/seed-golfathlete-db.js
cudom: P.S. To answer why I did not use id in the first place, it was because id was not mapping correctly to MongoDB's _id in the database. I still need to look into this a bit more.

cudom:
@tomasz-sf  @misiek-sf I just committed my changes based on your feedback. Thanks again!

test/app.spec.js
```diff
@@ -0,0 +1,187 @@
...
+
+const request = require('supertest')
+const chai = require('chai')
+const expect = chai.expect
+const mongoose = require('mongoose')
+const app = require('../src/app')
+
+describe('app', () => {
+  let id
+  let createdAt
```
patrick-sf, line 11: Can you please explain the reasoning for declaring `createdAt` here and with let?

src/controllers/athlete-controller.js
```diff
@@ -0,0 +1,77 @@
...
+    res.status(200).json(util.fromMongoDb(athlete))
+  } catch (err) {
+    return next(err)
+  }
+}
+exports.createAthlete = async (req, res, next) => {
+  try {
+    // todo index ghinNumber to avoid collection scan
+    const result = await Athlete.findOne({ ghinNumber: req.body.ghinNumber })
+    if (result && result.ghinNumber && result.ghinNumber === req.body.ghinNumber) {
```
patrick-sf: If you foresee this syntax being repeated across the code in the future, consider using lodash's get method to retrieve the value of a property which may or may not exist. Additionally, there is also a has method which returns a boolean whether the property exists or not.

test/app.spec.js
cudom: Prior to Tomasz comment about letting Mongoose handle the timestamps for updatedAt and createdAt, I captured it from the response so I could test later that it would not be improperly modified by an update (i.e. PUT). I made it a let here so that I can assign the value later but placed in a higher scope so it could be accessed by the test. That test was not in place but I have just added it. After setting timestamps equal to true in the mongoose schema, I am seeing behavior where the createdAt is being set to the same value as updatedAt. I am leaving this failing test in place so I take a look at it this weekend and troubleshoot what is going on. If you have any thoughts, any help would be appreciated :)

src/controllers/athlete-controller.js
cudom: Thanks. I will take a look at lodash's get and has method. This syntax is cumbersome.

cudom: I tested in repl and if (_.get(result, 'ghinNumber') === req.body.ghinNumber) is much cleaner. I haven't used lodash before but I can see it providing utility in a larger project.

PR #10, opened 2019-10-02
WIP: First pass at test

cbaron:
Wanted the team to get a look at some work I did yesterday.  Will be happy to iterate further as needed.


README.md
```diff
@@ -45,20 +45,20 @@ If you have not been specifically asked, you may pick either `Implementation Pat
...
 ### Basic Requirements
 
-  - Use Node.js `LTS` and any framework of your choice.
-  - Use any persistence store. NoSQL DB is preferred.
-  - Write concise and clear commit messages.
-  - Write clear **documentation** on how it has been designed and how to run the code.
+- Use Node.js `LTS` and any framework of your choice.
+- Use any persistence store. NoSQL DB is preferred.
+- Write concise and clear commit messages.
+- Write clear **documentation** on how it has been designed and how to run the code.
```
cbaron, line 51: I am using VSCode with an automatic lint on write.  I should probably add an eslint config to match your standards.

config/development.js
```diff
@@ -0,0 +1,14 @@
+const DYNAMODB_ENDPOINT = "http://localhost:8000";
```
cbaron, line 1: this should be dynamic based on the serverless.yml config

src/handler.ts
```diff
@@ -0,0 +1,192 @@
...
+type Container = {
+  config: {
+    DynamoDB: object;
+    stage: string;
+  };
+  dynamodb: DynamoDB;
+  client: DocumentClient;
+  TableName: string;
+};
+const retrieveContainer: () => Container = () => {
```
cbaron, line 18: The inversion of control pattern is something I like when dealing with endpoints as methods.

README.md
afaq-sf: Excellent. Follow-up question that:

How do you prefer to handle linting / auto-prettifying standards across multiple repositories?

For example, have you worked with any automation / CI hooks that check for linting or automatically prettify code based on config?

README.md
```diff
@@ -104,3 +103,39 @@ Fork this repo and send us a pull request when you think you are done. There is
 > What if I have a question?
 
 Create a new issue in this repo and we will respond and get back to you quickly.
+
+### Documentation
+
+Currently, the backend API service can be run on a local environment by running `yarn start`. This command uses [serverless](https://serverless.com), [serverless.yml](https://github.com/cbaron/node-backend-test/blob/master/serverless.yml), and the plugins referenced inside serverless.yml to emulate
+a local environment for developing against aws lambda methods.
```
afaq-sf: In order to get this running offline, I had to run install dynamodb libraries.

This is the error I encountered: serverless/serverless#3247

Not sure if that is a required setup step worth documenting.

config/development.js
afaq-sf: Good point. 
So, building on that, how would you change this code to handle that dynamic configuration injection?

serverless.yml
```diff
@@ -0,0 +1,83 @@
...
+      migrate: true
+provider:
+  name: aws
+  runtime: nodejs10.x
+  region: us-east-2
+  stage: ${opt:stage,'development'}
+  environment:
+    STAGE: ${opt:stage,'development'}
+
+functions:
```
afaq-sf, line 24: Have you used a serverless function's `environment` key to inject non-secret configurations?

src/handler.ts
afaq-sf: Can you elaborate a bit more on the abstraction cost and value that inversion of control provides here to the engineer/team working on this code?

src/handler.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  assert(user, "User Not Found");
+
+  return {
+    statusCode: 200,
+    body: JSON.stringify(user)
+  };
+};
+
+export const readUser: APIGatewayProxyHandler = async event => {
```
afaq-sf, line 55: We tend to keep each lambda path + method as a separate module.

Do you see any concerns with having a single handler file with multiple exported functions?

src/handler.ts
```diff
@@ -0,0 +1,195 @@
...
+  try {
+    const userId = event.pathParameters!.userId;
+    assert(userId, "Invalid request");
+    const container = retrieveContainer();
+
+    const response = await respondWithUser(userId, container);
+    return response;
+  } catch (e) {
+    return {
+      statusCode: 500,
```
afaq-sf, line 65: Awesome to see error handling incorporated in the handler functions.

1. What could be some pitfalls with defaulting to 500 error message for all unexpected errors?
2. If we wanted to inject a 3rd party error reporting tool such as Sentry, via its js SDK, how would the code inside the `catch` change?

src/handler.ts
```diff
@@ -0,0 +1,195 @@
...
+};
+
+export const readUser: APIGatewayProxyHandler = async event => {
+  try {
+    const userId = event.pathParameters!.userId;
+    assert(userId, "Invalid request");
+    const container = retrieveContainer();
+
+    const response = await respondWithUser(userId, container);
+    return response;
```
afaq-sf, line 62: I am curious of your thoughts on adding a return directly on line 60 to avoid the assignment to `response`.

src/handler.ts
```diff
@@ -0,0 +1,195 @@
...
+};
+
+export const deleteUser: APIGatewayProxyHandler = async event => {
+  try {
+    const userId = event.pathParameters!.userId;
+    assert(userId, "Invalid request");
+    const container = retrieveContainer();
+
+    const user = await getUser(userId, container);
+    assert(user, "User Not Found");
```
afaq-sf, line 173: If a user is not found for a delete operation, could it be considered a success? Why or why not?

src/handler.ts
```diff
@@ -0,0 +1,195 @@
...
+
+export const createUser: APIGatewayProxyHandler = async event => {
+  try {
+    assert(event.body, "Invalid request");
+    const { name, dob, address, description } = JSON.parse(
+      event.body as string
+    );
+
+    // ensure all fields are truthy
+    assert(name && dob && address && description, "Invalid body");
```
afaq-sf, line 79: Could you please describe how you would want to extend this to handle field-specific validation? 

For example: 
- description length limit
- address and DOB format

src/handler.ts
```diff
@@ -0,0 +1,195 @@
...
+        ...memo,
+        [`:${field}`]: body[field]
+      }),
+      { [":updatedAt"]: now }
+    );
+
+    const dynamoParams = {
+      TableName,
+      Key: { id: userId },
+      UpdateExpression: `SET ${updateExpressionList.join(", ")}`,
```
afaq-sf, line 149: Awesome work here.

Are you familiar with any alternative approaches to updating/upserting dynamo records that do not involve using expressions?

webpack.config.js
```diff
@@ -0,0 +1,26 @@
+const path = require("path");
+const slsw = require("serverless-webpack");
```
afaq-sf, line 2: What led you to use webpack for this use-case?

src/handler.ts
```diff
@@ -0,0 +1,195 @@
...
+    const { TableName, client } = container;
+
+    const dynamoParams = {
+      TableName,
+      Key: {
+        id: userId
+      }
+    };
+
+    await client.delete(dynamoParams).promise();
```
afaq-sf, line 184: If future requirements made it necessary to make this endpoint eventually consistent rather than awaiting the deletion from dynamo, how would you approach it?

afaq-sf (commented):
@cbaron Thank you for the awesome WIP submission.

I have left some initial questions on the current-state. I would appreciate your thoughts on those questions before any other work.

As for next iteration, it would be helpful to see:
1. Documentation of a plan on how CI/CD should be constructed for this code using any industry standard provider (GitHub Actions, Azure DevOps, CircleCI, etc.). No need to implement it unless time allows.
2. Any one of the Bonus Requirements listed in the README **OR** a simple front-end UI that communicates with this API and allows a user to be created and updated. Please feel free to use any frameworks or tools of your preference for the UI.

src/handler.ts
```diff
@@ -0,0 +1,195 @@
+import * as assert from "assert";
+import { APIGatewayProxyHandler } from "aws-lambda";
+import { DynamoDB } from "aws-sdk";
+import * as uuid from "uuid/v4";
+import getConfig from "./getConfig";
+import { DocumentClient } from "aws-sdk/clients/dynamodb";
+
+type Container = {
+  config: {
+    DynamoDB: object;
```
alex-sf: Can you touch on why you typed this as a generic object rather than a specific interface?

src/handler.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  return {
+    statusCode: 200,
+    body: JSON.stringify(user)
+  };
+};
+
+export const readUser: APIGatewayProxyHandler = async event => {
+  try {
+    const userId = event.pathParameters!.userId;
```
alex-sf: Im curious as to why you chose to do a non-null assertion vs typing the event?

README.md
cbaron: > How do you prefer to handle linting / auto-prettifying standards across multiple repositories?

For the most part, I care that standards exist, and that code cannot be put into master unless it has passed an auto linter.  I have been on a team that attempted to use a npm package to import into various repos for this purpose.  In our case, it wasn't seamless, but I think this is a fine approach.  On the other hand I don't think its a problem to have each repo's linting managed by the source maintainers -- especially if repos tend to vary in stack.

>For example, have you worked with any automation / CI hooks that check for linting or automatically prettify code based on config?

I have used Travis CI to run linters, tests before allowing to merge to master.  For a side project earlier this year I used CircleCI to do something similar.

README.md
cbaron: Its definitely worth documenting 😅 -- I think editing the README while I develop could be a useful way to catch this myself.  Peer review is also a good way 😄 

I will update.

config/development.js
cbaron: You can define environment variables in serverless based on stage, function, or use globals, so we could use a `process.env.DYNAMODB_ENDPOINT` to get this per environment.  We could probably get rid of `config/development` in favor of a general config and utilize serverless further.

serverless.yml
cbaron: Yes, I have.  I have also used serverless + aws secrets integration to deploy lambdas to staging/production VPCs.

My first goal was to get a local environment running with integration tests without thinking too much about deploying to staging or prod.  In the past I have done it manually from my cli using a `~/.aws/credentials` file.  I actually haven't personally done an aws deployment from CI/CD pipeline, but I am confident I could figure it out.

src/handler.ts
cbaron: Glad you called this out.  In the last typescript project I worked on, it was for a client who's tech lead had the final say on what code got merged.  This person insisted that we use `type` instead of `interface`, so this definition is the result of my haste.  I will change to interface as they are more flexible.

src/handler.ts
cbaron: As the implementation grows, it is likely that the `retrieveContainer` method will spend cpu cycles on things it may not necessarily need to fetch for the specific request.  The tradeoff for this is that an engineer working on an endpoint will generally not have to think about gathering any resources needed to implement a solution.

I have used IoC when working on an application deployed via docker.  After some thought, it is probably something to be avoided in lambda at scale.

src/handler.ts
cbaron: Having everything bundled into a single file is going to slow down lambda responses when they are brought into memory.  I will move into separate methods if you'd like.

src/handler.ts
cbaron: Good question.  I think it was either trying to be quick to make the type checker happy, or being overly sensitive to the Superformula team potentially not liking the verbosity of typescript?

I'm not actually sure of the best way to type `event.pathParameters`.  Event is an `APIGatewayProxyEvent` found in the `aws-lambda` package.

So I can do something like: 

```typescript
(event.pathParameters as APIGatewayProxyEvent["pathParameters"])
```

But event.pathParameters can still be `null`.

Maybe something like:

```typescript
type PathParameters = { userId: string };
const userId = (event.pathParameters as PathParameters).userId;
```

would be how I would iterate on this.  Curious how you would handle.

src/handler.ts
cbaron: That is how I originally implemented.  I _think_ that exceptions in `respondWithUser` were not being caught when I did this -- will follow up and confirm.

src/handler.ts
cbaron: 1.  Having everything return 500 makes it harder for clients to act appropriately.  To improve things we could either add more info to our assertions which allow this method to return more detailed error messages for clients, or use a utility method to do the same thing.

2.  We would want to log as much information as possible. The time of the exception, information about the request, the stack trace, etc would be passed to the reporting tool.

src/handler.ts
cbaron: I have used the [class-validator](https://www.npmjs.com/package/class-validator) package for this in a somewhat recent project, which I could see using again.  It does seem like classes have fallen out of favor in the javascript community.

Maybe I would try something like [dynamoose](https://dynamoosejs.com) which has validation support out of the box.  (I haven't tried this module).

I have worked a little with [dynaflow](https://github.com/VICEMedia/dynaflow) and while it doesn't have out of the box validation support, it deals with a lot of data very nicely and validate methods could be plugged in.

It is tough to say exactly what I would do without a roadmap of the service and its clients.

src/handler.ts
wprater: (I just edited to add syntax highlighting to the code blocks)

src/handler.ts
cbaron: Using AWS's AppSync service, I spent time writing somewhat reusable [mapping templates](https://docs.aws.amazon.com/appsync/latest/devguide/resolver-mapping-template-reference-dynamodb.html) to interface with DynamoDB.  I don't think this route is as ugly as one might first think :)

Hmm, come to think of it, I do not think I have any experience utilizing a DynamoDB orm.  Dynaflow doesn't really give one any syntactic sugar with regard to expressions.  https://github.com/clarkie/dynogels is something else worth investigating if I had no other input.  Curious what the team currently uses.

src/handler.ts
cbaron: Now I must go see what Roy Fielding said about this.

It could definitely be considered a success depending on context, but if I would lean toward sending a more informative status code -- perhaps a 404, so that the client can make an informed decision based on the response.

src/handler.ts
cbaron: Hmm, I am not sure if I understand you correctly, or if I am over thinking things.

If we don't want to await the deletion from dynamo, then we can remove the `await` on this line.  Obviously this has repercussions if there is an error.  Let me know if there is more to explore here.

webpack.config.js
cbaron: Because I have used it in the past to bundle typescript / javascript.  Curious how the team approaches it.

src/handler.ts
cbaron: If I changed lines 60/61 here to simply:

```typescript
return respondWithUser(userId, container)
```

Then any exceptions thrown in respondWithUser will not be caught.  I wouldn't be opposed to wrapping portions of the implementation/functions in order to catch and respond with more detailed error messages.

src/handler.ts
alex-sf: Thanks for the response on this! I definitely understand that typescript is not the quickest to prototype with so there's always plenty of `!` and `as any` whenever I test things. But there are a few ways you could handle this:

1) In the variable declaration (preferred method)
```ts
export const readUser: APIGatewayProxyHandler = async (event: APIGatewayProxyEvent) => {...}
```
2) Through nested traversal (which is what you had brought up)
```ts
const pathParam = event.pathParameters as APIGatewayProxyEvent["pathParameters"];
const userId = pathParam.userId;
```

Although, It looks like neither of these aren going to solve your problem because looking at their code there's always going to be a null for that param:
```ts
export interface APIGatewayProxyEvent {
    ...
    pathParameters: { [name: string]: string } | null;
    ...
}
```
The real issue is that I think the typing is not super great in the `aws-lambda` package. They should really extend their interface as a constructor that can take in variables to create the proper definitions. That way you can inform the interface that you have defined `pathParameters`.

So I think the answer to this would have to be the custom definition solution you went with:
```ts
type PathParameters = { userId: string };
const userId = (event.pathParameters as PathParameters).userId;
```

Or a pull request against the library but that's a little out of scope here. 

Awesome work 👍

cbaron:
@wprater , @afaq-sf , @alex-sf -- Just added a very simple React application which demonstrates network calls to the API.

Its relatively quick and dirty with regard to code organization.  Normally if I'm doing a monorepo, I would

- include tsconfig, linting, webpack configs in a root level `packages` directory (as local modules) where yarn [workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) can be utilized such that all applications within the repo can access these packages as if they existed on npm.
- break out each application into its own folder within `/apps`

To get the front end up and running, run `yarn start-client`.
I changed the backend to run via `yarn start-api`.

wprater:
Thanks for your recent update, @cbaron .  We will be checking them out soon!

wprater:
Thank you for your submission, we will continue offline.

PR #12, opened 2019-10-08
node-backend-test implementation

greenreign:
@wprater @afaq-sf 

Let me know if you have questions or would like to discuss this.

-zak


afaq-sf:
Thank you for this submission, Zach.
We will review it soon and share any questions on the PR.

design-decisions.md
```diff
@@ -0,0 +1,28 @@
+# Design Decisions
+Assumptions, thoughts and decisions for current and future consideration.
+
+##  Technologies and frameworks
+- Typescript
+- TSLint
+- DynamoDb - The problem prefers it and it has good options for local development in a docker image or serverless plugin
+- Documentation - Rely on github for now but it would be nice to have polished api documentation
```
afaq-sf, line 8: What would polished API documentation for a real-world project?

Have you used any tools or services to auto-generate API specifications and documentation?

design-decisions.md
```diff
@@ -0,0 +1,28 @@
...
+- serverless-dynamodb-local - We have other options with dynamodb-local through [docker](https://hub.docker.com/r/amazon/dynamodb-local) which I've used recently.  But I really wanted to evaluate this library. Unfortunately it requires Java.  For me that's not a huge issue but something that could become an issue for larger teams.
+
+- Validation - Try to handle with Json Shema and API Gateway.  I was able to get validation working with API Gateway and a schema defined in serverless BUT there is an open issue in serverless-offline which means it will not work locally.  Thus I wasn't able to add tests for it.  See below for a more optimal long-term approach.  
+- Testing (mocha, tsnode, jest) - I chose jest with ts-jest.  It worked well enough.  I ran into an issue where my unit tests really weren't going to offer much value but integration tests would. Integration tests working with all the local tooling to replicate the aws serverless environment with a single command would certainly be possible and optimal but was taking too much time.  For now I will run two commands to run tests :(
```
afaq-sf, line 14: That is very fair considering the scope of the test.

> I ran into an issue where my unit tests really weren't going to offer much value but integration tests would.

How did you come to that conclusion? How do you tend to evaluate testing needs for your code?

design-decisions.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Future Considerations
+- Logging - I'm using pino which I've used before.  It's fast and defaults to json which works well with cloudwatch and other logging aggregators.  We can enhance logging to record information about each request.  We could then combine logging with cloudwatch to cheaply get some traction on a few items below (monitoring and analytics).
```
afaq-sf, line 17: Very glad to see you called this out. Especially the benefits of using off-the-shelf logging middleware like pino.

Have you worked with or created centralized logging infrastructure that ships from CloudWatch to 3rd party providers like AppDynamics/DataDog/etc?

design-decisions.md
```diff
@@ -0,0 +1,28 @@
...
+- Monitoring - We would want to introduce monitoring and alarms for latency, call volume, errors etc
```
afaq-sf, line 18: How would you go about getting a centralized look at the health of the overall service using AWS services?

For example, similar metrics are available in API Gateway Dashboard, CloudFront Distribution Dashboard, ALB/Target group consoles, Lambda built-in dashboards and CloudWatch offerings. Which do you tend to use and why?

design-decisions.md
```diff
@@ -0,0 +1,28 @@
...
+- Security - I have used AWS WAF to protect against DDoS, general bots, throttling, etc in the past.
+- Analytics - This may or may not be important.  Something to consider while evaluating the needs of the client
```
afaq-sf, line 20: When it has been necessary, can you give us an example of one or two tools you have used for analytics on domain operations within the service?

package.json
```diff
@@ -0,0 +1,39 @@
...
+    "ts-jest": "^24.1.0",
+    "tslint": "^5.20.0",
+    "typescript": "^3.6.3"
+  },
+  "dependencies": {
+    "moment": "^2.24.0",
+    "pino": "^5.13.4",
+    "uuid": "^3.3.3"
+  }
+}
```
afaq-sf, line 39: Do you tend to use linters / code formatters that would handle line endings outside of tslint?

serverless.yml
```diff
@@ -0,0 +1,112 @@
+
+
+service: users-api
+
+provider:
+  name: aws
+  runtime: nodejs10.x
+  iamRoleStatements:
```
afaq-sf, line 8: What are your thoughts on best practices around role reuse?
Do you tend to give each Lambda its distinct permissions or create IAM roles that can be reused across Lambdas?

serverless.yml
```diff
@@ -0,0 +1,112 @@
...
+        - dynamodb:GetItem
+        - dynamodb:PutItem
+        - dynamodb:UpdateItem
+        - dynamodb:DeleteItem
+      Resource:
+        - arn:aws:dynamodb:${opt:region, self:provider.region}:*:table/users
+
+
+package:
+  exclude:
```
afaq-sf, line 21: Excellent. Great to see you are removing unnecessary overhead to lighten the Lambdas.

serverless.yml
```diff
@@ -0,0 +1,112 @@
...
+
+  createUser:
+    handler: dist/index.createUser
+    events:
+      - http:
+          path: users
+          method: post
+          request:
+            schema:
+              application/json: ${file(dist/schema/userCreate.json)}
```
afaq-sf, line 67: Awesome. Nice to see API GW schema being leveraged.

How would you handle custom error messages for the schemas that are handled by API GW? Any benefit to moving this validation to within the Lambda code itself?

src/core/UserController.ts
```diff
@@ -0,0 +1,47 @@
+import { User } from '../model/User';
+import { UserCreate } from '../model/UserCreate';
+import { UserUpdate } from '../model/UserUpdate';
+import { UserDataAccess } from './UserDataAccess';
+
+export class UserController {
```
afaq-sf, line 6: What value does this Controller abstraction provide over allowing Lambdas direct access to Data Access layer?

src/index.ts
```diff
@@ -0,0 +1,63 @@
...
+    const user: User = await userController.getUser(id);
+    return {
+      statusCode: 200,
+      body: JSON.stringify(user),
+    };
+  }, event);
+};
+
+export const createUser: Handler = async (event: any): Promise<APIGatewayProxyResult> => {
+  return await handle(async () => {
```
afaq-sf, line 24: Have you in the past seen issues / limitations with handling exceptions via a wrapper rather than having them explicitly defined in the handler code?

afaq-sf (commented):
Hi @zaksesti,

Thank you again for your thoughtful submission. I enjoyed reviewing it.

I have left some questions that would be helpful to get your perspective on.

design-decisions.md
greenreign: > What would polished API documentation for a real-world project?
> 
> Have you used any tools or services to auto-generate API specifications and documentation?

@afaq-sf This depends on the needs and priorities.  I've used swagger in the past.  The OAS spec is supported by many tools and frameworks.  You can use it to generate validation, docs and models.  Like I mentioned a single source of truth.  But the auto generated docs for OAS in swagger are just OK?  If your API is a large part of your business value then I'd suggest you have something even more polished.  Like Stripe's API docs.  Optimally this is still something that's auto generated from a single source of truth.   But in this case you may need another tool/framework or something homegrown.   At amazon their was custom tooling in place to generate models and docs from a json-like domain specific language (spec).

design-decisions.md
greenreign: > How did you come to that conclusion? How do you tend to evaluate testing needs for your code?

@afaq-sf Good question.  Lets look at the opportunities for unit testing here.
- `UserController.ts`  I'm looking for "units" where some value would be gained by testing them individually.  At first there was some value to testing this.  But it eventually became mostly a pure delegator with little or no actual business logic.  It's value at all can be debated separately.  At that point I have to mock the data access layer.  Then I'm writing tests like
```
expect(mockedDataAccess.createUser).toBeCalledWith(exactUserCreateObject);
```
This offers very little value.  Confirming that I've passed objects through unmodified.  Mocking objects is not always trivial.  It could have some very small value to prevent regressions but it's value is unlikely to ever be realized against the cost of maintaining these tests through refactors and enahancements.  

With an integration test I get all the coverage with much less cost.  This is a balancing act.  

Another example is
- ` ExceptionMapper.ts`  
I get 100% coverage for this class in my integration tests but this is an example where Unit Tests would have very little cost.  Also it may not be obvious as I'm changing my integration tests that I'm expected to cover this class.  `ExceptionMapper` is a good example, especially as it gets more complicated, of a class we should be unit testing.  It will be easy to prevent regressions and enhance this class.  

- `UserDataAccess`:  This class gets it's coverage from the integration tests.  Unit testing will have the overhead of mocking the DynamoDb DocumentClient.  This isn't trivial but this class does actually contain some business logic.  We could test that it errors when expected.  We could test that id's and audit columns (updated/created) are populated when expected.  I've found that these types of tests have value early on in a project and then you begin the repeat the same functionality over and over.  I usually add them early and decide potentially to remove them later when it makes sense and you have the coverage from other tests.

Tests are extremely valuable.  They allow us to move faster with confidence.  Writing testable code inherently produces more maintainable and expandable code.  However, every line of code you add has a cost in the short and long term.  You need to balance these investments

design-decisions.md
greenreign: > Have you worked with or created centralized logging infrastructure that ships from CloudWatch to 3rd party providers like AppDynamics/DataDog/etc?

@afaq-sf I have not specifically created a stream from CloudWatch to a 3rd party provider such as DataDog etc.  I am aware of the value of a polished all-encompassing dashboard.  I know that it's fairly straight forward to front your monitoring with 3rd party tools.  My requirements have always fallen short of those needs.  We created monitoring for internal engineers.  In those cases I choose the smallest barrier to entry with pure cloudwatch/aws solutions. 

It's not to say that I didn't consider more extensive implementations but I chose to build smaller iterations and wait for those problems to prioritize themselves.

design-decisions.md
greenreign: > How would you go about getting a centralized look at the health of the overall service using AWS services?
> 
> For example, similar metrics are available in API Gateway Dashboard, CloudFront Distribution Dashboard, ALB/Target group consoles, Lambda built-in dashboards and CloudWatch offerings. Which do you tend to use and why?

@afaq-sf   Ok, this is a great one.  I truly value the ability to see the health of all services/infrastructure in a glimpse.  A single dashboard.  A single screen if possible.  This is the holy grail.  I advocated for this at Amazon and it resulted in TVs in our team areas with rolling dashboards of our services (request volume, latency, cache hits etc) and our deployment pipelines.  We loved it.

I have not gotten to a point in other projects where I've reproduced that experience to that extent but if a project were to prioritize such transparency I'd love to tackle it.  

I have used cloudwatch dashboards to display graphs and alarms for all the data you've mentioned above.  This is fairly simple and something you would create in your infrastructure as code.  If we wanted to get more polished or more detailed then I think you could start looking into third party tools.  DataDog, DynaTrace, etc.

design-decisions.md
greenreign: > When it has been necessary, can you give us an example of one or two tools you have used for analytics on domain operations within the service?

@afaq-sf Sure, at Onu I used Keen.io to record usage stats.  I added a layer into my rest framework's request/response chain.  It would record certain events.  For example, we had a client who wanted to know which sales persons were logging in and which subset of API calls they were making and how often.  So we could record this to Keen and then we would query the data back for reporting.  Similarly we would record total data transferred to/from our media service by upstream services and their users.  This allowed us to tier pricing if needed for users who had heavy media usage.  

In general (before Keen) I also used a cloudwatch reporter for our java rest framework (dropwizard).  This allowed us to easily record request/response metrics to cloudwatch.  We used it mostly for monitoring but it could easily be used for some short term analytics.  

At NS8 we have a more homegrown analytics system for fraud assessment.  Using javascript and batching data through an api to mongodb.  This is subsequently queried with aggregation to assess session/order risk.

serverless.yml
greenreign: > Excellent. Great to see you are removing unnecessary overhead to lighten the Lambdas.

@afaq-sf Thank you.  By default serverless will not include dev dependencies which is nice.  I can also make aws-sdk a dev dependency because lambda includes those in the environment automatically.  The unfortunate thing is because of the way serverless uses the globs I can't exclude everything and then only include a few select items.  That would be more future proof but at quick glance this did the trick for now.

serverless.yml
greenreign: > What are your thoughts on best practices around role reuse?
> Do you tend to give each Lambda its distinct permissions or create IAM roles that can be reused across Lambdas?

I think this is another balancing act depending on the requirements.  For maximum security you would break down roles as finely as possible and restrict access to only what was absolutely necessary to that piece of compute.  This comes with overhead however and could become difficult to manage.  In many scenarios there may not be any risk in giving the `createUser` lambda the same permissions as the `updateUser`, `getUser` and `deleteUser` lambda.  Therefore you can reduce overhead by just using the same role.  

However I could envision scenarios where a user model may have a separate datastore(table) for NPPI data (social security number etc.).  This data may only be read perhaps by roles with elevated privileges ( a very specific service or admin).  This would seem an obvious reason to segregate the `readUsers` iam role from the `readUsersNppi` iam role.

serverless.yml
greenreign: > How would you handle custom error messages for the schemas that are handled by API GW? 

I'll be completely honest here and admit that I'm not 100% sure.  This is an area where I think there has been an explosion of growth lately.  Even with API gateway I think there are several ways to validate so I'm not sure if perhaps your OAS spec could instruct API gateway on it's validation and even messaging with replacement params.  That would take some more proofing.

> Any benefit to moving this validation to within the Lambda code itself?  

Certainly if you move the validation inside lambda then you have more flexibility.  If you run into a wall with API Gateway where you are unable to do something (like you said custom messages, conditional validation, pure business logic) then  you can certainly accomplish it in the code.   This also ensures that ALL error handling is in one place and can be made to be consistent.  

This may be a bit outside the bounds of this question but I had, at one point at NS8, implemented a generator.  It would take my Swagger 2 (OAS 3 was not available yet) and on commit of my swagger repo would generate models along with some manipulation of the swagger spec to make the JsonShema compliant to use with my model validation.  The idea was that it would eventually generate clients as well (node, php, java, python).  I do believe this is probably even easier now with OAS 3 and supported tooling and would likely be worth looking into.

src/core/UserController.ts
greenreign: > What value does this Controller abstraction provide over allowing Lambdas direct access to Data Access layer?

@afaq-sf Yeah, it really isn't doing anything at this point. It is however providing a facade to index.ts which is abstracting the AWS (APIGateway/Lambda) implementations from the rest of the application. If you wanted to pull this application out of lambda tomorrow and deploy it on ECS you'd only need to replace index.ts with a server (express, restify, fastify, nestjs) routing layer.  You could leave `UserController` in place.  This may seem like a thin argument as you could claim the `UserDataAccess` is doing that as well.  But as this application grows the controller is likely to gain some business logic or at least delegate to more downstream dependencies.  In that case it will be easier to replace the lambda layer and test this layer on it's own.  

Perhaps premature optimization here but experience tells me this will be needed soon.

src/index.ts
greenreign: > Have you in the past seen issues / limitations with handling exceptions via a wrapper rather than having them explicitly defined in the handler code?

@afaq-sf   I don't believe so.  Or, at least, the value gained in not repeating the wrapper code has always outweighed the edge cases?  I could imagine a very specific error which would like to be handled only by the method which called it.  I think I would consider that more a code smell than an argument for handing errors in the handler.  Although there are always exceptions. 

All rest frameworks I've used provide default error handlers (wrappers) like this and allow you to create a mapping for any types of errors you don't wish to be handled by the default.  This adds great value knowing I can't miss an error and all errors will have a predictable result to the client.  But I can still override that behavior if I need to.  

Since I'm not using a framework here I created a simple wrapper.  However at the onset of a project I would consider whether it may be valuable to use a framework instead of raw handlers.  There are lambda-express plugins.  Similar with fastify and nestjs I believe.  These could offer real value in error handling, validation, middleware, dependency injection etc.

package.json
greenreign: Wow, I have to imagine there are plenty of solutions.  vscode/webstorm settings.  Maybe eslint has a linter or plugin for json.  I'm sure one of the prettifier's would handle it.  Ashamed to say I don't know offhand.  Vscode was taking care of the formatting and validation.  Kinda surprised it didn't nag me on the last newline.  I used webstorm previously and was happy to try out vscode here but I hadn't run across problems with the newlines before.

wprater:
Thank you for your submission.  We will continue offline.

PR #14, opened 2019-10-18
Feature/cloud native solution

ivanovj:
A PR with my solution of the cloud-native path. Please note - while the CRUD functionality is in place as REST endpoints, that commit reflects a work in progress. There's a quite a bit that can be done to improve it - testing, docs, GraphQL and so on. Yet, I hope you guys could see where I am going with it - including architecture, configuration, implementation details, etc.

As always any and all your comments are truly welcomed. Will be glad to help!

### DIFF
--- a/solution-jivanov/.gitignore
+++ b/solution-jivanov/.gitignore
@@ -1,0 +1,12 @@
+# package directories
+node_modules
+jspm_packages
+
+# Serverless directories
+.serverless
+
+# IDE folders
+.idea
+
+# Other folders
+coverage/
--- a/solution-jivanov/README.md
+++ b/solution-jivanov/README.md
@@ -70,7 +70,7 @@ example:\
 **Important note:** each endpoint is protected with an API key. It will be generated upon deployment and displayed on-screen (look at the deployment section above). It must be provided as part of the request header "x-api-key: API_KEY_HERE"
 
 
-RESPONSE MODEL (for all of the above operations):
+**Response model** (for all of the above operations):
 ```
 {
     message: "success or error message",
@@ -83,7 +83,7 @@ RESPONSE MODEL (for all of the above operations):
  
 ---
 
-###Filesystem structure
+### Filesystem structure
 - `/` - root folder contains config files and handlers of the endpoints
 - `/serverless.yml` - main microservices config file (IaC)
 - `/dynamodb.yml` - CloudFormation IaC directives to provision the DB  
@@ -91,7 +91,7 @@ RESPONSE MODEL (for all of the above operations):
 - `services/` - contains all services classes
 
 
-###Additional notes:
+### Additional notes:
 1. Utilizing the isomorphic ValidationJS library for the input validations. That would allow to re-use the same on the frontend too (and potentially leverage the full-stack javascript).
 2. Field validations are kept at minimum at this time (due to none specified in the requirements), but could be easily amended thanks to the above mentioned library.
 3. 'Name' is the only mandatory field (when creating a new record); that is to avoid creating empty user records.
--- a/solution-jivanov/README.md
+++ b/solution-jivanov/README.md
@@ -1,0 +1,102 @@
+### OVERVIEW
+This is a serverless implementation of CRUD for DynamoDB on NodeJS, exposed as REST endpoints via API gateway. 
+Computing layer is AWS Lambda. Utilizing serverless framework as orchestration tool, leveraging Infrastructure-as-code (IaC).
+
+---
+
+### KISS
+My overarching approach to implementations is [KISS](https://en.wikipedia.org/wiki/KISS_principle). Keep it as simple as possible, as straightforward as possible. Both architecture-wise as
+well as code-wise. Don't try to shove every trick from the book in there. Instead, just "KISS" it. 
+Emphasize on readability, maintenability and extensibility. Simple as that!
+
+---
+
+### HOW TO INSTALL
+Thanks to CloudFormation and our tooling (Serverless) we are able to write all infrastructure as code (IaC). That in turn really simplifies our installation task:
+
+First, make sure you have [Serverless Framework](https://serverless.com/) installed and properly configured (refer to [their get started page](https://serverless.com/framework/docs/getting-started/) as needed)
+
+After that to install all dependencies please run the (all popular) command of:
+
+`npm install`
+
+
+### HOW TO RUN TESTS
+
+We are utilizing [Jest](https://jestjs.io/) as testing framework. It is already setup in our project. To execute all the automated tests simply run:
+
+`npm run test`
+
+
+
+### HOW TO DEPLOY
+The whole project is architected with multiple environments in mind (dev, staging, prod, other as needed). Currently "dev" is setup as default. For added flexibility, as well as to keep it simple (and as advised in the [serverless docs](https://serverless.com/framework/docs/providers/aws/guide/credentials/)) each environment correspond to a separate AWS profile. Dev environment corresponds to "dev" profile. For that reason please make sure you have an AWS profile named "dev" set up, before you attempt a deploy with:
+
+
+`serverless deploy -v`
+
+Thanks to the Infrastuture-as-code (IaC) the above command will provision all the needed infrastructure, deploy all the code, setup endpoints,
+ usage plans, database, anything and everything.\
+ It will return the AWS secret key as well as the actual endpoints URL. Please note all of those, you will need them to make it work in the next step.  
+
+### HOW TO USE
+Upon successfull deployment there will be 4 REST endpoints created as outlined below. Each one handles a particular CRUD operation as follows:
+
+**POST /users** - performs CREATE operation i.e. creates a new user record. Returns the newly created user ID (uuid of the user)
+
+example:\
+`curl -X POST -H "x-api-key: KEY_HERE" https://AWS_URL_HERE/dev/users --data '{"name": "TEST User 8", "dob":"1970-07-17", "address": "Mandalay Bay Blvd 17", "description":"This is a test user"}'`
+
+--
+
+**GET /users/{id}** - performs READ operation i.e. retrieves user record details by the given user ID (uuid)
+
+`curl -X GET -H "x-api-key: KEY_HERE" https://AWS_URL_HERE/dev/users/USER_UUID_HERE`
+
+--
+
+**POST /users/{id}** - performs UPDATE operation i.e. updates user record data of the given user ID (uuid)
+
+`curl -X POST -H "x-api-key: KEY_HERE" https://AWS_URL_HERE/dev/users/USER_UUID_HERE --data '{"dob":"1980-08-18"}'`
+
+--
+
+**DELETE /users/{id}** - performs DELETE operation i.e. removes user record with the given user ID (uuid)
+
+`curl -X DELETE -H "x-api-key: KEY_HERE" https://AWS_URL_HERE/dev/users/USER_UUID_HERE`
+
+--
+
+**Important note:** each endpoint is protected with an API key. It will be generated upon deployment and displayed on-screen (look at the deployment section above). It must be provided as part of the request header "x-api-key: API_KEY_HERE"
+
+
+RESPONSE MODEL (for all of the above operations):
+```
+{
+    message: "success or error message",
+    itemId: "the UUID of the user the request has been made with or the generated one upon CREATE",
+    data: "optional data returned by READ which will contain the full user record,
+    timestamp: "UTC timestamp of the operation"
+ }
+ ```
+ 
+ 
+---
+
+###Filesystem structure
+- `/` - root folder contains config files and handlers of the endpoints
+- `/serverless.yml` - main microservices config file (IaC)
+- `/dynamodb.yml` - CloudFormation IaC directives to provision the DB  
+- `models/` - contains all model files/classes
+- `services/` - contains all services classes
+
+
+###Additional notes:
+1. Utilizing the isomorphic ValidationJS library for the input validations. That would allow to re-use the same on the frontend too (and potentially leverage the full-stack javascript).
+2. Field validations are kept at minimum at this time (due to none specified in the requirements), but could be easily amended thanks to the above mentioned library.
+3. 'Name' is the only mandatory field (when creating a new record); that is to avoid creating empty user records.
+4. All service errors are returned as 4xx; 5xx are reserved for infrastructure (AWS) errors; HTTP code of 200 is OK/success.
+5. All returned error messages are prefixed with "ERROR:" for easier identification.
+6. There's a sample usage plan configured in the serverless.yml file that implements throttling and quota settings. 
+7. For added security each endpoint is protected with an API key generated upon deployment.
+8. The configuration is architected with multiple environments in mind (dev, staging, prod, other as needed); currently "dev" is setup as default.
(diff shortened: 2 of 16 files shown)


solution-jivanov/README.md
```diff
@@ -0,0 +1,102 @@
+### OVERVIEW
+This is a serverless implementation of CRUD for DynamoDB on NodeJS, exposed as REST endpoints via API gateway. 
+Computing layer is AWS Lambda. Utilizing serverless framework as orchestration tool, leveraging Infrastructure-as-code (IaC).
+
+---
+
+### KISS
+My overarching approach to implementations is [KISS](https://en.wikipedia.org/wiki/KISS_principle). Keep it as simple as possible, as straightforward as possible. Both architecture-wise as
+well as code-wise. Don't try to shove every trick from the book in there. Instead, just "KISS" it. 
+Emphasize on readability, maintenability and extensibility. Simple as that!
```
afaq-sf, line 10: What mechanisms (documentation, SDLC processes, and check-points, etc.) would you want as feedback loops for reinforcing these principles over the course of a project?

solution-jivanov/README.md
```diff
@@ -0,0 +1,102 @@
...
+### Filesystem structure
+- `/` - root folder contains config files and handlers of the endpoints
+- `/serverless.yml` - main microservices config file (IaC)
+- `/dynamodb.yml` - CloudFormation IaC directives to provision the DB  
+- `models/` - contains all model files/classes
+- `services/` - contains all services classes
+
+
+### Additional notes:
+1. Utilizing the isomorphic ValidationJS library for the input validations. That would allow to re-use the same on the frontend too (and potentially leverage the full-stack javascript).
```
afaq-sf, line 95: > re-use the same on the frontend too

Do you see any issues with this isomorphic approach for multi-channel API consumers? 

For example, if we need to consume this API on Kiosks, B2B APIs, mobile apps, etc. Would that present challenges / adjustments to this approach?

solution-jivanov/README.md
```diff
@@ -0,0 +1,102 @@
...
+Thanks to CloudFormation and our tooling (Serverless) we are able to write all infrastructure as code (IaC). That in turn really simplifies our installation task:
+
+First, make sure you have [Serverless Framework](https://serverless.com/) installed and properly configured (refer to [their get started page](https://serverless.com/framework/docs/getting-started/) as needed)
+
+After that to install all dependencies please run the (all popular) command of:
+
+`npm install`
+
+
+### HOW TO RUN TESTS
```
afaq-sf, line 24: Looking beyond its current simplicity, what sort of testing would you want to put thought into for Enterprise-grade APIs?

Any tools/services you have experience with that you rely on?

solution-jivanov/.gitignore
```diff
@@ -0,0 +1,12 @@
...
+jspm_packages
+
+# Serverless directories
+.serverless
+
+# IDE folders
+.idea
+
+# Other folders
+coverage/
```
afaq-sf, line 12: Do you tend to use linting/formatting tools standardizing line-endings?

solution-jivanov/dynamodb.yml
```diff
@@ -0,0 +1,15 @@
...
+    Properties:
+      TableName: ${self:custom.usersTableName}
+      AttributeDefinitions:
+      - AttributeName: id
+        AttributeType: S
+      KeySchema:
+      - AttributeName: id
+        KeyType: HASH
+      ProvisionedThroughput:
+        ReadCapacityUnits: 1
```
afaq-sf, line 14: If capacity provisioning is defined in Serverless spec, how would you approach dynamic autoscale needs for Dynamo tables?

solution-jivanov/dynamodb.yml
```diff
@@ -0,0 +1,15 @@
+Resources:
+
+  UsersTable:
+    Type: AWS::DynamoDB::Table
```
afaq-sf, line 4: If we need to front this Dynamo table with read/write-through cache layer for scaling needs, how would you approach that?

solution-jivanov/lib/base.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    // returns true if s is a string, false otherwise
+    // based on: https://stackoverflow.com/questions/4059147/check-if-a-variable-is-a-string (3rd answer)
+    static isString(s) {
+        return Object.prototype.toString.call(s) === "[object String]";
+    }
+
+    // returns true if f is a function, false otherwise
+    static isFunction(f) {
```
afaq-sf, line 28: Can you help me understand the value of writing low-level operations like `isNumeric`, `isString`, and `isFunction` into the base class?

And are there any alternative methods that you know of to remove this level of abstraction and implementation?

solution-jivanov/models/users.js
```diff
@@ -0,0 +1,283 @@
...
+            minimum: 2, // and must be at least 2 chars long
+            message: 'must be at least 2 chars long'
+        },
+    },
+};
+const VALIDATION_OPTIONS = {
+    format: 'flat', // return "flat" message format i.e. a single string
+};
+
+const USERS_TABLE_NAME = process.env.USERS_TABLE_NAME;
```
afaq-sf, line 73: Would you see any value in using destructuring syntax here?

solution-jivanov/models/users.js
```diff
@@ -0,0 +1,283 @@
...
+
+// a map to translate input params to DynamoDB ones if needed
+const dynamoDbKeys = new Map ([
+    ['name', '#userName'],
+]);
+
+
+class Users extends Model {
```
afaq-sf, line 80: It seems like this Users model class, specifically, is responsible for input validation, data access layer communication details, and generic abstraction of user methods.

I would like to understand your approach on scope-limiting OOP for Models.

solution-jivanov/lib/model.js
```diff
@@ -0,0 +1,31 @@
...
+// Model class - base model class
+// contains functionality that is/can be common for all models
+
+const Base = require('./base.js');
+
+const AWS = require('aws-sdk');
+
+let dynamoDb;
+
+class Model extends Base{
```
afaq-sf, line 11: I've noticed you took a relatively abstracted OOP approach for the lambda code to leverage.

How do you come to that design decision and what were the alternatives you considered?

solution-jivanov/serverless.yml
```diff
@@ -0,0 +1,148 @@
...
+        path: users/{id}
+        method: post
+        integration: lambda-proxy
+        cors: true
+        private: true
+
+  delete:
+    handler: users.delete
+    timeout: 10
+    memorySize: 1024
```
afaq-sf, line 125: How do you tend to evaluate Lambda resource provisioning?

solution-jivanov/serverless.yml
```diff
@@ -0,0 +1,148 @@
...
+    - http:
+        path: users/{id}
+        method: post
+        integration: lambda-proxy
+        cors: true
+        private: true
+
+  delete:
+    handler: users.delete
+    timeout: 10
```
afaq-sf, line 124: What led to the decision to setting 10s as the timeout?

solution-jivanov/users.js
```diff
@@ -0,0 +1,132 @@
+
+// Require needed libraries and modules
+const Service = require('./services/users.js');
+
+module.exports.create = async (event, context) => {
+    console.log("users/create context= ", context);
+    console.log("users/create event= ", event);
+
+    let httpCode = 200; // default is success
```
afaq-sf, line 6: What would happen if an error occurs before the `try` block?

afaq-sf (commented):
Hi @ivanovj 

Thank you for your submission. I've left some comments and questions. If you have time, it would help to get your thoughts on them.

solution-jivanov/README.md
ivanovj: Oh, man, love that question, thank you! It is quite big a topic too. But let me try to provide some bullet points at least:

- initial training - every new team member goes through some period of initial training where those principles are explained, demonstrated and discussed

- documentation - we need readily available docs that (again) explain and showcase those principles with various examples; team member should be able to easily refer to those docs.

- coding standards - well documented, (again) explained and documented through the above

- linting tool to automate and enforce (at least some) of those standards

- code reviews - very important; ideally we would like to review every piece that goes to production; that would be a great checkpoint for reinforcing those

- building the dev team with the understanding that code quality is a team responsibility, not just a single person one; each member should participate and put effort towards that. 

- tools like SonarQube to perform static code analysis as well as cyclomatic complexity one (very helpful to reinforce KISS and writing maintainable code in general)

and we can probably think of more items...Big topic...Let me know if you would like to discuss further.

solution-jivanov/README.md
ivanovj: The idea behind utilizing isomorphic libraries is to empower consumption, to make it easier. There's generally a problem with code being reused on the back and front end, even for full-stack JS projects. And isomorphic libraries attempt to amend that, enable the front and backend to share, re-use code pieces (unify validation in this case). 
Thinking about it further Validate.JS runs on ECMAScript 5.1 i.e. anything that supports ECMAScript 5.1 would be able to run it. And that is pretty wide, nowadays anything that can run Javascript would support ES5...
But even it something does not, the even better part is that validation rules are in declarative format - as JSONs (which could potentially be shared with the various consumers). And virtually everything can read and parse a JSON nowadays. Again, talking about enabling consumption...  
For anyone that can not (for whatever reason to) read JSON nor to run ES5, then they would implement validation on their end as with any other API that their would consume...

solution-jivanov/README.md
ivanovj: Integration, functional, regression testing suites...I've worked with Selenium before, would be really happy to hear what you guys use and prefer. Always interested to learn!

solution-jivanov/.gitignore
ivanovj: Oooh, yes, I've seen "wars" between senior devs on whether to use single or double quotes :) Certainly, lining is very useful and can catch a lot of issues. My IDE of choice (WebStorm/PhpStorm) has built-in linting and I rely on it every day. If you guys have any particular rules on endings, etc. please let me know, will be happy to adjust my IDE.

solution-jivanov/dynamodb.yml
ivanovj: I put 1 here primarily for cost concerns, have not planned for autoscaling on this particular task. But that certainly can be done, there are (at least) 2 ways to do it: 

- We can define the table to be "on-demand" (and AWS will happily accommodate our loads). 

- OR we can add auto-scaling to our DynamoDB table (Serverless even has an autoscaling plugin which allows us to control the auto-scaling policy as IaC). 

The primary difference between the 2 is "control". The latter gives us more control over the auto-scaling.

solution-jivanov/dynamodb.yml
ivanovj: DAX is probably a good choice. And it can be provisioned via CloudFormation too i.e. it is fairly straightforward IaC as well.

solution-jivanov/README.md
ivanovj: a couple of additional items (that we've been doing) in that regard:
- pair programming - very important for explaining and reinforcing the major principles as well as going into greater details and particular use cases.
- regular presentation from various dev team members - showcasing best practices, new technologies, open-ended discussions and so much more. 
- a motivated team lead to execute on those - my goal is always to help and empower the team; both individually and as a whole.

solution-jivanov/lib/base.js
ivanovj: There's no value in terms of the particular service, other than me trying to showcase understanding of (a bit) more complicated architectures, abstraction, inheritance, etc. :)

It is also a good point you make about the "...*low-level operations*..."...Unfortunately still not available as built-in features of the language...We should probably submit a proposal for the ESNext...

"*And are there any alternative methods that you know of to remove this level of abstraction and implementation?*"
Certainly, quite a bit of libraries have those built-in (lodash, etc.). ValidateJS does too. We could use it straight in the users model (where we load it anyways). Again, no real need of having them here...

solution-jivanov/models/users.js
ivanovj: In this particular case...mmm...don't see anything obvious really...process.env is always an object...and yes, with destructuring you could set a default, but you could do that without it too...at the same time this is easier to read than destructuring...please let me know if I am missing anything...

solution-jivanov/models/users.js
ivanovj: Certainly, let me try to elaborate please. My approach here is KISS. 
As per wikipedia a model is "The central component of the pattern. It is the application's dynamic data structure, independent of the user interface.[5] It directly manages the data, logic and rules of the application." [Reference here](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller)

Everything that directly works with the Users data would go there. 
Anything else would go to a different model - for example if the users have roles which are defined (and perhaps stored) somewhere else, then that would be a separate model (UserRoles)

I would not try to force scope-limiting a model unless it becomes too big/unreadable. In the case of this particular model, we could split it in like "UserValidations", "UserDatabase" and so on.

But again, in general, I find it much easier, much more efficient to work with a single file (assuming it is not too big of course), it is just so convenient to have everything under your finger tips...

solution-jivanov/lib/model.js
ivanovj: Oh, yes, as mentioned above - just trying to showcase understanding of more stuff (abstraction, classes, inheritance, etc.). Not really needed here.

The good thing about microservices (compared to a monolith) is that we can make them as simple (or as complex) as needed, completely independently from the other microservices. If a service can benefit from that level of abstraction - then sure, let's use it. If it does not need it - no need to have it either. "KISS" in action :)

For that particular service - we certainly do not need it. A simple service and a model classes would have been enough.

solution-jivanov/serverless.yml
ivanovj: It depends on the use case really. Is it CPU consuming? does it make external API calls? do we plan on calling it from the front-end(i.e. would like to return faster)? and so on...

If there are no particular requirements I would usually start with MemorySize of 1024Mb. That is the Serverless framework default too. From my experience it provides good CPU power (Lambda allocates CPU power proportional to the memory size) as well as enough memory for most tasks. It is also very easy to compute the cost due to that round number.

solution-jivanov/serverless.yml
ivanovj: Again - good, round number to start with. Enough to cover the cold-starts and some. Returns fast enough in case of service failure (actual timeout). Easy to compute the cost due to that round number as well. 
Please check additional considerations in the above answer; I would adjust the timeout based on the particular use case too.

solution-jivanov/users.js
ivanovj: if an error occurs before the "try" block then it will be thrown and that service (AWS Lambda) will end with 500 error. That would be immediately caught by our monitoring service (and/or log ingestion one), we will be alerted and can investigate right away (and/or roll back the latest deployment). We should be alerted for all 500s for any of our services.
Of course, we should do our best to avoid such a situation and be very careful of what we put before the "try" block, if anything at all. We prefer to always catch all errors and display a polite/user-friendly message in case of an error.
In this particular case there's nothing that can really throw (due to the setup event is always an object; not because those statements don't throw in general), but that's a good point you make, let me try to minimize that risk further, just pushed an update about that. 

Also, we should probably have a "soft" deployments strategy in place, in order to reduce the impact of issues in general. Or at least blue-green deployments. A proper deployment strategy would help us reduce the risk of issues, downtime and overall impact. As well as recover faster.

wprater:
Thank you for your submission.

PR #15, opened 2019-10-24
node-backend-test-cj-o

mmonkey:
Here is my submission for the cloud-native node-backend-test. Please let me know if you have any questions, comments or concerns. I will follow up via email with live stack information.

Thanks!


.eslintrc.json
```diff
@@ -0,0 +1,32 @@
...
+    "object-curly-spacing": ["error", "never"],
+    "prefer-promise-reject-errors": "off",
+    "semi": ["error", "always"],
+    "space-before-function-paren": ["error", {
+      "anonymous": "always",
+      "named": "never",
+      "asyncArrow": "always"
+    }]
+  }
+}
```
afaq-sf, line 32: How do you recommend enforcing linting across the team to assure all of the rules are applied before code is merged to master?

src/lambdas/users-delete.js
```diff
@@ -0,0 +1,57 @@
...
+ * @apiParam {String} id User's unique identifier.
+ *
+ * @apiSuccessExample Success-Response
+ * HTTP/1.1 204 No Content
+ *
+ * @apiUse ResourceNotFoundException
+ * @apiUse Exception
+ */
+export async function handler(event, context, callback) {
+	console.log(event);
```
codin-sf, line 29: Since your logs are being sent to AWS Cloudwatch, how would you go about implementing a pattern which is consistent across logging statements? Is there a transporter with a pattern you would suggest?

src/lambdas/users-delete.js
```diff
@@ -0,0 +1,57 @@
...
+
+	try {
+		let user = new User();
+		user.id = event.params.path.id;
+
+		const client = new DyanamoDB();
+		const mapper = new DataMapper({client});
+
+		user = await mapper.get(user).catch(err => {
+			throw (err.name && err.name === 'ItemNotFoundException') ? new ResourceNotFoundException('User', user.id) : err;
```
codin-sf, line 39: Why not have an Instance of an object rather than passing back a string for an exception?

bin/build-templates.js
```diff
@@ -0,0 +1,35 @@
...
+	'{{DOCS_RELEASE_BUCKET}}': process.env.DOCS_RELEASE_BUCKET,
+	'{{LAMBDAS_RELEASE_BUCKET}}': process.env.LAMBDAS_RELEASE_BUCKET
+};
+
+/**
+ * Read in each .template from the templates directory.
+ * Then rewrite all replacements with the appropriate value.
+ * Save rewritten .template files to dist/templates directory.
+ */
+mkdir(outputPath)
```
codin-sf, line 22: Where would a failure be caught?

bin/build-templates.js
```diff
@@ -0,0 +1,35 @@
+const {promisify} = require('util');
```
afaq-sf, line 1: How would you plug these build scripts when using CI/CD service providers (CircleCI, AWS Codebuild, Azure Pipelines, etc.) that tend to be single-file custom-spec oriented (yaml + bash)?

For example, would the CI/CD spec file just run these JS files to build/deploy? 
Or would they be separate commands outside these JS scripts to build/deploy these applications?

bin/release-templates.js
```diff
@@ -0,0 +1,5 @@
+const {exec} = require('child_process');
+require('dotenv').config();
+
+const cmd = exec(`aws s3 cp dist/templates/ s3://${process.env.TEMPLATES_RELEASE_BUCKET}/ --recursive --exclude "*" --include "*.template"`);
+cmd.stderr.on('data', data => console.log(`> ${data}`));
```
afaq-sf, line 5: What value would this JS script provide over using the same AWS CLI command directly in npm script or as discreet bash scripts?

src/exceptions/invalid-attribute.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+export default class InvalidAttributeException extends Exception {
+	/**
+	 * InvalidAttribute constructor
+	 * @param {string} detail
+	 */
+	constructor(detail) {
+		super(detail);
+
+		this.status = 422;
```
afaq-sf, line 28: How would you abstract this abstracted exception handling strategy across Lambdas in different repositories?
Any alternative approaches you can recommend?

src/lambdas/deploy-api-docs.js
```diff
@@ -0,0 +1,96 @@
...
+		const sourceBucket = event.ResourceProperties.SourceBucket;
+		const rewriteFiles = event.ResourceProperties.RewriteFiles || [];
+
+		try {
+			// Copy files to destination bucket
+			const params = {
+				Bucket: sourceBucket,
+				Prefix: 'docs'
+			};
+			await s3.listObjects(params).promise().then(response => {
```
afaq-sf: Would you recommend any alternative control-flow here to avoid then/catch clauses? Why or why not?

For example, perhaps awaiting the response from `listObjects().promise()` and assigning to a variable.

src/lambdas/users-create.js
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @apiUse InvalidAttributeException
+ * @apiUse Exception
+ */
+export async function handler(event, context, callback) {
+	console.log(event);
+
+	try {
+		let user = new User(event.body);
+		await user.validate();
```
afaq-sf, line 84: Love the minimal lambda code which focuses on flow-control more than inline logic.

Do you see any value in making these `validate()` methods more explicit rather than completely abstracted as it is here?

src/lambdas/users-create.js
```diff
@@ -0,0 +1,107 @@
...
+		await user.address.validate();
+
+		const client = new DyanamoDB();
+		const mapper = new DataMapper({client});
+
+		user = await mapper.put(user);
+
+		// Clear the CloudFront cache
+		const lambda = new Lambda();
+		const params = {
```
afaq-sf, line 94: Do you see any alternative approaches to avoid invoking this lambda directly?

What would happen if this cache invalidation failed but the preceding logic had all passed?

src/lambdas/users-delete.js
```diff
@@ -0,0 +1,57 @@
...
+		const params = {
+			FunctionName: `${process.env.STACK_NAME}-CloudFrontCreateInvalidation`,
+			InvokeArgs: JSON.stringify({
+				paths: ['/api/v1/users', '/api/v1/users?*', `/api/v1/users/${user.id}`]
+			})
+		};
+		await lambda.invokeAsync(params).promise();
+	} catch (err) {
+		console.log(err);
+		callback(new ExceptionHandler(err));
```
afaq-sf, line 55: Can you help me understand the reason for using the lambda `callback` method vs returning exception directly?

src/models/user.js
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * DynamoDB Schema configuration
+ * See: https://github.com/awslabs/dynamodb-data-mapper-js/tree/master/packages/dynamodb-data-mapper
+ */
+Object.defineProperties(User.prototype, {
+	[DynamoDbTable]: {
+		value: `${process.env.STACK_NAME}-Users`
+	},
+	[DynamoDbSchema]: {
```
afaq-sf, line 89: :+1: on keeping schema as code.

test/exceptions/test-handler.js
```diff
@@ -0,0 +1,44 @@
...
+			]);
+			assert.ok(handler.errors[0] instanceof Exception);
+			assert.strictEqual(handler.errors[0].detail, 'Custom error message.');
+			assert.ok(handler.errors[1] instanceof InvalidAttributeException);
+			assert.strictEqual(handler.errors[1].detail, 'Invalid attribute message.');
+			assert.ok(handler.errors[2] instanceof ResourceNotFoundException);
+			assert.strictEqual(handler.errors[2].detail, 'Resource with id \'unknown\' does not exist.');
+		});
+	});
+});
```
afaq-sf, line 52: What criteria do you use to decide which cases/scenarios are worth testing?

afaq-sf:
Thank you for your submission!

SF team members are leaving some questions/comments. Would be awesome to get your input on them as time allows.

README.md
```diff
@@ -1,106 +1,131 @@
...
+* Both the API and documentation site are delivered through the same CloudFront distribution.
+* The API _requires_ an https connection, where as the documentation site will _redirect_ the user to https automatically.
+* All _GET_ requests are cached. This includes both API and documentation requests.
+* The distribution will forward any querystring parameters onto the API (for _GET_ filtering).
 
-- The API should follow typical RESTful API design pattern.
-- The data should be saved in the DB.
-- Provide proper API documentation.
-- Proper error handling should be used.
+**ApiGateway**
```
afaq-sf, line 22: If this project needed to pivot away from REST and be a Graph API instead, how would you approach that?

For example: 
What tools/services would you use? 
Which aspects of your code would need to entirely change? 
Which aspects of your code could remain or be easily adapted?

src/lambdas/users-delete.js
mmonkey: A recommended approach is to introduce something like [winston](https://github.com/winstonjs/winston) for logging. Then defining CloudWatch patterns for errors or general logging would be more straight forward and consistent.

src/lambdas/users-delete.js
mmonkey: In this case, mapper.get() will throw two types of errors. I feel that the requesting party should only be responsible for providing a valid id, for this endpoint, in order to make changes to the User with that id. If they didn't provide a valid id, then mapper.get() will respond with an error object with the name property set as `ItemNotFoundException`. If that is the case, we want to respond with a 404, and therefore a `ResourceNotFoundException`. However, in any other case, an error thrown is not the responsiblity of the requester, and could have any number of reasons. For that case, we want to pass the error through to be logged to CloudWatch, and returns a generic 500 Error, and we will be alerted via the alarms.

In both cases, an object is used to handle the errors, and I only use string comparison to determine which error is being returned from the library.

.eslintrc.json
mmonkey: Currently the linter will run automatically during the build and the release npm scripts, using the prebuild hook. A git pre-commit hook could also be used to trigger the linter to also enforce the action. The package [pre-commit](https://www.npmjs.com/package/pre-commit) comes to mind for this.

bin/build-templates.js
mmonkey: I like to see as much pairity across development, continouous integration and production as possible. I have limited experience with CI/CD, but I do know that several of them use docker to build and deploy. It is likely that I'd have to break out parts of the build process to support the provider.

bin/release-templates.js
mmonkey: Your recommendation of just using an npm script or bash script is definitely my first choice, however, I had issues testing a Windows machine using that method. In the case of these release scripts, they are calling an environment variable set by the developer, and I was unable to get cross-env or dotenv to pass the proper info into the aws cli command on Windows.

src/exceptions/invalid-attribute.js
mmonkey: Two approaches come to mind.

1. The logic could be broken out into it's own node package, and be called like any other library.
2. The logic could run on a Lambda Layer and be shared among the functions that require it.

src/lambdas/users-create.js
mmonkey: I went back and forth on this. I can see the use-case of validating specific properties, or changing validation requirements based on the context, like requiring a state only if a city is provided. I think adding an optional field whitelist would be the next step.

src/lambdas/users-delete.js
mmonkey: I had to choose between API gateway mapping errors with message templates, or to handle everything in the Lambda function. Because I decided to handle the messaging through API Gateway, I found that I had to use the callback in order to ensure the correct format and status codes.

test/exceptions/test-handler.js
mmonkey: I tried to test all cases that have an outcome of passing or failing, as well as "forks in the road". For example:

1. Calling a library with the same parameters every time should have a pass or fail outcome.
2. Changing the filter parameter on users-list is what I consider a "fork in the road", and should have a few different outcomes.

README.md
mmonkey: Unfortunately I don't have a great answer for this as I don't have experience beyond some light reading about Graph APIs or GraphQL. However, according to the [Serverless blog](https://serverless.com/blog/make-serverless-graphql-api-using-lambda-dynamodb/), I can continue using API Gateway and DynamoDB. It looks like what changes the most is the models and lambda functions. I think I can define GraphQL Schemas and Mutations along side the current schema definitions and go from there.

bin/build-templates.js
mmonkey: Good call, I'll push an update to address this.

src/lambdas/deploy-api-docs.js
mmonkey: Absolutely right, this function does stick out a bit for it's flow-control differences from the others, I think I can clean up most of it. I'll push an update to address this.

src/lambdas/users-create.js
mmonkey: The reason that this lives in a separate Lambda function in the first place is because passing the `DISTRIBUTION_ID` environment variable to the API functions caused a circular dependency. There are several ways to handle this, the ones I was deciding between where:
1. Invoke a separate function that could live outside the circular dependency.
2. Load the distribution id from the Systems Manager Parameter Store.

I also debated whether or not throw the error if cache invalidation fails, and thinking through it again, it should probably just log an move on. Update incoming.

wprater:
Thanks for your submission, we will be reaching out to you shortly!

PR #17, opened 2019-12-06
Submission for Cloud Native

GrahamEvans31:
Thanks for the opportunity and taking the time to speak with me on the phone and now review my submission.

Look forward to discussing my work and hope to have a next step and join the team.

Regards,
Graham


README.md
```diff
@@ -1,106 +1,43 @@
...
 
-## Q&A
+### Logging, Monitoring and alarming
 
-> Where should I send back the result when I'm done?
+Throughout the API I have made use of the console for logging, using info/error/warn where appropriate.  These logs will naturally flow into CloudWatch with the corresponding log group for each API and stage it is in (prod/dev).  Using metric filters we can pattern match and emit a value that would be alarmable with a CloudWatch alarm.  I would use a low sev/high sev pair, where a low sev may not require action but is noted that there is a potential issue and a high sev indicating that action is required immediately.  All cloud watch alarms can be modeled in IaC in the serverless.yml resource files.  This keeps any manual setup to a minimum for a developer.
 
-Fork this repo and send us a pull request when you think you are done. There is no deadline for this task unless otherwise noted to you directly.
+Since we are using managed services, each service like Lambda/API Gateway/DynamoDB already emit their own sets of metrics that can be alarmed on.  This requires fare less augmentation of the code to emit your own metrics.
```
afaq-sf, line 39: @GrahamEvans31 How would you recommend creating dashboards / alarms based on these aws-managed service metrics?

For example, if the team wanted to create a CloudWatch dashboard for this service via infrastructure-as-code, how could that be done via this repo?

README.md
```diff
@@ -1,106 +1,43 @@
...
 
-> What if I have a question?
+### Filter Feature
 
-Create a new issue in this repo and we will respond and get back to you quickly.
+I implemented a basic query on the name of the user, provided a query parameter with "name" is present in the request.  I wanted to call out a few things about this solution, given DynamoDB is naturally a key/value store and any query or table scan operations are expensive it is paramount when building a feature like this to understand the exact types of read queries the application must respond to.  Without clarity here, operations could be very expensive ($) and correct LSI or GSI need to be available to support these.  Also, if multiple attributes were to be queried for, a composite key of concatenated attributes is a great solution to build an index off of.  The last thing I wanted to mention about the method of implementation is that I chose query parameters for bookmarkability, if I were to have included it as a POST with a body, we would not have that functionality which can be great for front end applications.
```
afaq-sf, line 43: 👍

libs/dynamodb-lib.js
```diff
@@ -0,0 +1,10 @@
+import AWS from "aws-sdk";
+
+/**
+ * Bootstrap and provide method to call DynamoDB.
+ */
+export function call(action, params) {
+  const dynamoDb = new AWS.DynamoDB.DocumentClient();
+
+  return dynamoDb[action](params).promise();
```
afaq-sf, line 9: Hypothetically, speaking, how would you recommend extending this code if the dynamoDB client began name-spacing features/versions under object properties?

For example, if the `someAction` method existed that also needed to be accessed here: `dynamoDb.v2.someAction(params).promise`

libs/response-lib.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * Handle a successfuly API call return case.
```
afaq-sf, line 2: If more Lambdas existed across git repos, how would you recommend re-using these common helpers?

libs/response-lib.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+export function success(statusCode, body) {
+  return buildResponse(statusCode, body);
+}
+
+/**
+ * Handle a failed API call return case.
+ */
+export function failure(statusCode, body) {
```
afaq-sf, line 11: Imagine a scenario where the client requires each HTTP failure status code to have a custom `customMessage` property. The value of this property must be retrieved from a standalone API.

How could that mechanism be incorporated to `response-lib`?

afaq-sf (commented):
@GrahamEvans31 Thank you for submitting this PR.

I have left some questions that would be awesome to get your thoughts on.

README.md
GrahamEvans31: Haven't personally done this before, but some light reading shows that CloudFormation supports dashboarding (in the console), albeit in a painful manner with specifying the exact grid coordinates in pixels...  Some quick research shows that there are plugins available for Serverless to specify dashboard under the custom section in serverless.yml.  I'd want to explore a few of these and settle on one that allows us to describe these in our IaC and avoid any manual dashboarding.  Internal amazon was manual dashboarding, these dashboards frequently became out of touch with the underlying metrics, if metrics are added or changed, this is not a great situation to be in.

libs/response-lib.js
GrahamEvans31: We could create a node package, that would be depended on across the repos to give a uniform integration with DynamoDB.  These helpers were just a basic approach I took in this assignment.

libs/dynamodb-lib.js
GrahamEvans31: If the caller needed to explicitly select a versioned API I'd look to add in a version parameter or some agreed upon best practice to allow the caller to select the version they prefer.  In a more ideal situation, where the caller may not need to know or care, it could be encapsulated inside the method.

I think what is important is looking at future migration, do we want our callers to have to migrate their code to a new version (as they are specifying which version they are calling) or do we want to control what version is being called so as to respect a basic interface contract the clients are calling.  This may be a case by case situation.

libs/response-lib.js
GrahamEvans31: As this code is laid out, its very basic and strict with respect to the code/body parameters.  I'd hate to have a function that required a large amount of parameters, but I'd also be wary of taking in an object where all the properties need to be inspected and then matched to the response.  There is likely a happy medium here, adding in a customMessage parameter doesn't seem like to much scope creep for this function but it could quickly lead to some bad code smells that becomes difficult to maintain if others followed suit.  A custom message could also have a null default value that wouldn't require it be provided to keep it clean when others don't have this requirement.

GrahamEvans31 (commented):
Responded, I wasn't sure if "resolved" just meant that I'm passing ball back to you if there are additional comments please reopen the thread or add another comment.  I think I interpreted it as acknowledged.

create.js
```diff
@@ -0,0 +1,31 @@
...
+    if (errors) {
+      return failure(HttpStatus.BAD_REQUEST, {errors : errors});
+    }
+
+    let user = new User(dynamoDbLib);
+    try {
+      const result = await user.create(data);
+      return success(HttpStatus.CREATED, result);
+    } catch (e) {
+      return failure(HttpStatus.INTERNAL_SERVER_ERROR, {});
```
tomasz-sf, line 25: I'd suggest not hiding the actual error. At least logging it would be a good idea but also returning at least some information to the client is preferable.

delete.js
```diff
@@ -0,0 +1,38 @@
...
+  try {
+    if (!('pathParameters' in event)) {
+      return failure(HttpStatus.UNPROCESSABLE_ENTITY, {error : "ID Path Parameter required."});
+    }
+    const id = event.pathParameters.id;
+
+    if (!id) {
+      return failure(HttpStatus.UNPROCESSABLE_ENTITY, {error : "ID Path Parameter required."});
+    } else if (!id.match(UUID_V1)) {
+      throw { message : "Required UUID V1 format." };
```
tomasz-sf, line 20: Can you explain why did you choose to throw plain object instead of `Error`.
I'm also not sure what's the pattern here. Sometimes you are using `throw` and sometimes `return failure`

delete.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    let user = new User(dynamoDbLib);
+    try {
+      const exists = await user.read(id);
+      if (!exists) {
+        return failure(HttpStatus.NOT_FOUND, {});
```
tomasz-sf, line 27: I'd say repeated passing of empty object is a little bit wasteful. My suggestion would be either handling undefined body in the success/failure or just setting default value on the body argument.

model/__mocks__/user.js
```diff
@@ -0,0 +1,77 @@
...
+  async create(user) {
+    if (user.name === "EXCEPTION") {
+      throw "Internal Server Error";
+    }
+    return Promise.resolve(user);
+  }
+
+  async read(id) {
+    if (id === "f18f46e0-16d7-11ea-8260-1541765a4531") {
+      return Promise.resolve({
```
tomasz-sf, line 13: IIRC you don't need to `Promise.resolve` here because everything returned from async method is automatically a Promise

update.js
```diff
@@ -0,0 +1,43 @@
...
+    const id = event.pathParameters.id;
+    const data = JSON.parse(event.body);
+    const errors = validate(data, userValidationCriteria);
+
+    if (!id) {
+      return failure(HttpStatus.UNPROCESSABLE_ENTITY, {error : "ID Path Parameter required."});
+    } else if (!id.match(UUID_V1)) {
+      throw {message : "Required UUID V1 format."};
+    } else if (errors) {
+      return failure(HttpStatus.BAD_REQUEST, {errors : errors});
```
tomasz-sf, line 25: `{errors : errors}` can be shortened to just `{ errors }`

update.js
```diff
@@ -0,0 +1,43 @@
...
+        return failure(HttpStatus.NOT_FOUND, {});
+      }
+      const result = await user.update(id, data);
+      return success(HttpStatus.OK, result);
+    } catch (e) {
+      return failure(HttpStatus.INTERNAL_SERVER_ERROR, {});
+    }
+  } catch (e) {
+    console.info("Invalid request, body: " + event.body + " pathParameters: " + JSON.stringify(event.pathParameters));
+    return failure(HttpStatus.UNPROCESSABLE_ENTITY, {});
```
tomasz-sf, line 41: Can you explain the decision behind making `422` a default error?

read.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+    const id = event.pathParameters.id;
+
+    if (!id) {
+      return failure(HttpStatus.UNPROCESSABLE_ENTITY, {error : "ID Path Parameter required."});
+    } else if (!id.match(UUID_V1)) {
+      throw {message : "Required UUID V1 format."};
+    }
+
+    let user = new User(dynamoDbLib);
```
tomasz-sf, line 23: I'm not sure what this pattern of passing `dynamoDbLib` is achieving. Looks like you could just import it inside of User (or any other such classes).

create.js
GrahamEvans31: The internal error is logged in the user object (in this code it really could only come from the dynamodb call),  but I would not agree on returning this to the client.  This is an internal server error, it is not advised to indicate what is happening inside of your service to clients.  This just gives unnecessary information about the underlying service and its dependencies, it could be a security risk.

delete.js
GrahamEvans31: I don't think I chose the plain object over error, I'd likely want all of them to be the standard Error object.  When I was writing the unit tests it became required from the JEST framework to be the Error object, so I adjusted some but not all.

So when I throw on line 20, it is being caught and then returned under the UNPROCESSABLE_ENTITY, I'm just using my try/catch flow to reuse that failure.  For instance, I don't want to tell the client that we expect UUID V1 as thats our privileged information but it falls under the same error response code as maligned JSON.  What you return to clients is important to assist them in calling your API but its also sensitive information about our application.

delete.js
GrahamEvans31: Wasteful in terms of?  I'm not sure I see how this impacts performance in a meaningful way, I understand an empty parameter is not really of any use, probably could set this as a default value then it wouldn't be required.

model/__mocks__/user.js
GrahamEvans31: You could be right, I didn't try that and I think I was just being explicit in my mock as to what I wanted returned.

read.js
GrahamEvans31: This is to mock the functions from dynamoDbLib for testability, its dependency injection.  It is simpler from the unit test side of things to mock a function then to mock an import.

update.js
GrahamEvans31: Its not a default, this requires an exception to be thrown to be returned.

update.js
GrahamEvans31: Without the errors key its not as clear in the response that they are errors, this was my mentality.  But I think it could work without as well.

create.js
tmarciszewski: I agree that giving end-user too much specific information about internal errors is not a good idea. I was not happy with catching error and not doing anything with it apart from just sending generic internal error.

GrahamEvans31:
@tomasz-sf Sorry didn't see your questions until today, left a response under each, resolve if you feel they are answered or if my answer isn't clear let me know.

delete.js
tmarciszewski: It probably has none or negligible impact on performance. Just repeating this empty object over and over again instead of adding default to the argument seemed strange.

update.js
tmarciszewski: I understand it needs error to be thrown but this is a "cover-all" error handler. All kinds of errors could happen here. You chose to return 422 from it instead of the usual 500 and I wanted to know what was the reason for this.

tomasz-sf:
Heh. Now I've made a mess by commenting from my private account... Generally, I'm happy with your responses. Thank you :-)

create.js
GrahamEvans31: Yea I think I battled with not logging an error here as it would be duplicated in the event of a dynamodb failure from the user model logging, but this does expose it to not logging anything if there was a separate error.  So it likely would be best to not log in the user model and only log here which would give all the information in any event.  Then I could even remove the try/catch in the dynamodb calls

update.js
GrahamEvans31: The reason was simply that I was expecting the catch to be either of my throws in this actual segment of code.  Given that the user model does throw an error, this could lead to a cover all capturing a real 500 and returning the 422, so I see the deficiency here.  Even though, I do wrap the user model in its own try/catch, its hard to see how the remaining lines could trigger a error that isn't related to an issue with the request.

GrahamEvans31:
All good, thanks and left an additional comment on your 2nd round.

GrahamEvans31:
@afaq-sf @tomasz-sf 

Guess I couldn't quite show I was proficient in AWS services, not really sure how you arrived at that conclusion but good luck finding your more ideal candidate.

